PR #13, opened 2026-04-05
fix: add Helmet middleware to set security headers

Migz93:
## Summary
- Installs and applies the `helmet` npm package as the first Express middleware
- Removes the `X-Powered-By: Express` header from all responses
- Adds sensible default security headers (X-Content-Type-Options, X-Frame-Options, etc.)

## Notes
- Zero functional impact on the app
- Resolves Snyk finding: Information Exposure - X-Powered-By Header (MEDIUM)

### DIFF
--- a/src/server/app.ts
+++ b/src/server/app.ts
@@ -147,10 +147,17 @@ export function createApp(config: RuntimeConfig, scheduler?: JobScheduler) {
     contentSecurityPolicy: {
       directives: {
         ...helmet.contentSecurityPolicy.getDefaultDirectives(),
-        "connect-src": ["'self'", "https://plex.tv", "https://api.github.com"]
+        "connect-src": ["'self'", "https://plex.tv", "https://api.github.com"],
+        // Disable upgrade-insecure-requests: this app runs over plain HTTP in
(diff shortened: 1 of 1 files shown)


Migz93:
@codex we've made a fix to that but then claude has also found the following, what do you recommend?

Found something. User avatar URLs from Plex (friend.user.avatar, user.thumb) come in as absolute URLs from Plex's API. The getPlexImageSrc function only proxies paths starting with / — absolute URLs fall through and are loaded directly by the browser. Helmet's default img-src 'self' data: would block those.

Everything else is clean:

All poster/artwork images go through the /api/plex/image proxy (same-origin)
No external fonts or CDN scripts
index.html is entirely self-contained
The fix options for avatars:

Add https://plex.tv to img-src — but Plex avatars can also come from other CDNs (e.g. Gravatar, Imgur) depending on what users have set on their Plex accounts. Fragile.
Add img-src: 'self' data: https: — allows any HTTPS image source. Less strict but pragmatic for a self-hosted app where you can't enumerate all avatar CDNs.
Route avatar URLs through the existing server proxy — most secure, but requires extending buildTrustedPlexImageRequest to handle absolute Plex URLs.

PR #14, opened 2026-04-05
fix: address Snyk type validation and DOM XSS findings

Migz93:
## Summary
- Replace `as string` casts on `req.query["filter"]` and `req.query["search"]` with `typeof` guards — prevents silent type confusion if query params are arrays
- Add `typeof` guard on `payload.hostname` before calling `.trim()` — prevents runtime TypeError on malformed request bodies
- Validate `changelogRelease.html_url` starts with `https://github.com/` before using as `href` — eliminates Snyk DOM XSS finding

## Test plan
- [x] Log viewer still filters and searches correctly
- [x] Manual Plex configuration (hostname + port) still saves successfully
- [ ] Changelog modal "View on GitHub" link opens the correct release page

### DIFF
--- a/src/client/pages/Settings.tsx
+++ b/src/client/pages/Settings.tsx
@@ -851,7 +851,7 @@ function AboutTab() {
                 Close
               </button>
               <a
-                href={changelogRelease.html_url}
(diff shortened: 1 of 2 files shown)


Migz93:
@codex review

PR #29, opened 2026-04-08
fix: increase image proxy byte limit to cover Plex Gracenote poster art

### LINKED ISSUE #28: feat: replace image proxy with local image cache
## Summary

Currently all external images (movie/show poster art and user avatars) are proxied at request time through `/api/plex/image` and `/api/avatar`. This works but has downsides:

- Every image load hits the proxy, adding latency
- High-resolution CDN images (e.g. Plex Gracenote art can be 5+ MB) are fetched and streamed on every request
- The proxy endpoints add complexity and have been the source of bugs (e.g. 2 MB byte cap silently dropping large posters)

## Proposed approach

At sync time, download and resize each image to a web-appropriate resolution and store it on disk. Serve images directly as static files rather than proxying.

**Scope:**
- Poster art (movie & show thumbnails) — resize to something like 300×450px on ingest
- User avatars — download and store at original size (already small)
- Serve cached images from a local path (e.g. `/config/image-cache/`)
- Remove `/api/plex/image` and `/api/avatar` proxy endpoints once the cache covers all use cases

**To be decided / out of scope for initial implementation:**
- Cache invalidation / periodic refresh strategy (how often to re-fetch images, or trigger on sync)
- "Clear image cache" button in Settings UI
- What to show as a fallback while an image is being fetched for the first time

## Benefits

- Faster image load times (local disk vs. remote CDN fetch per request)
- No more size limit issues in the proxy
- Reduces outbound network traffic after initial sync
- Simpler client code — `getPlexImageSrc` can become a static path lookup

Migz93:
## Summary

- Plex Gracenote CDN images (used as movie poster thumbs) can be 5+ MB, exceeding the previous 2 MB cap in the avatar proxy
- TV show thumbs come from TMDB at ~1 MB so they loaded fine; movie posters were silently rejected with a 502 and showed no image at all
- Raises the cap to 20 MB and renames the constant to `MAX_IMAGE_PROXY_BYTES` to reflect that the endpoint now serves both user avatars and full-resolution poster art

## Test plan

- [ ] Movie posters (e.g. GOAT, Wuthering Heights) now load correctly on the Watchlists and Dashboard pages
- [ ] TV show posters continue to load as before
- [ ] User avatars continue to load as before

Closes #28 is the longer-term fix (local image cache replacing the proxy entirely).

### DIFF
--- a/src/server/app.ts
+++ b/src/server/app.ts
@@ -55,7 +55,10 @@ const ALLOWED_PLEX_IMAGE_QUERY_PARAMS = new Set(["width", "height", "minSize", "
 // (fc00::/7 = fc and fd prefixes), IPv6 link-local (fe80::/10).
 const PRIVATE_IP_RE =
   /^(127\.|10\.|172\.(1[6-9]|2\d|3[01])\.|192\.168\.|169\.254\.|::1$|\[::1\]|::ffff:|\[::ffff:|f[cd][0-9a-f]{2}:|\[f[cd][0-9a-f]{2}|fe80:|\[fe80:|localhost)/i;
-const MAX_AVATAR_BYTES = 2 * 1024 * 1024;
+// The image proxy serves both small user avatars (<500 KB) and full-resolution
+// Plex Gracenote poster art which can exceed 5 MB. 20 MB gives ample headroom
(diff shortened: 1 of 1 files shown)


Migz93:
@codex /review

PR #30, opened 2026-04-09
feat: expand Playwright test suite with feature and API tests

Migz93:
## Summary

- Adds 6 new spec files (~30 new tests) covering Dashboard UI, Watchlists filters/sorting, History filters, Users page structure, Settings tabs, and API smoke tests
- Fixes auth storageState cookie format — missing `expires` field caused the `request` fixture (used in API tests) to fail immediately
- Fixes a pre-existing strict mode violation in the Settings smoke test in `pages.spec.ts` (`getByRole("heading", { name: "Settings" })` was substring-matching `<h3>General Settings</h3>`)

## New test files

| File | What it covers |
|---|---|
| `dashboard.spec.ts` | Stat chips, filtered links, Recently Added section, Recent Syncs panel, Run Sync button |
| `watchlists.spec.ts` | Filter chips present, sort options present, URL param updates for each filter |
| `history.spec.ts` | Type/status filter buttons, page size select, URL param updates |
| `users.spec.ts` | Active section, Disabled accordion, Refresh Users button |
| `settings.spec.ts` | All 6 tabs present, tab URL navigation, tab content spot-checks |
| `api.spec.ts` | `/api/health`, `/api/auth/session`, `/api/dashboard` shape via `request` fixture |

## Test plan

- All 41 tests pass against a live instance (`npm run test:e2e`)
- All tests are read-only — nothing writes to Plex or modifies app state

### DIFF
--- a/.env.playwright.example
+++ b/.env.playwright.example
@@ -1,0 +1,2 @@
+# Copy this to .env.playwright and set your running instance URL
+BASE_URL=http://localhost:3000
--- a/.env.playwright.example
+++ b/.env.playwright.example
@@ -1,2 +1,11 @@
-# Copy this to .env.playwright and set your running instance URL
+# Copy this to .env.playwright and fill in both values.
+#
+# BASE_URL: URL of your running Hubarr instance
 BASE_URL=http://localhost:3000
+#
+# SESSION_COOKIE: your hubarr_session cookie value, grabbed from the browser.
+#   1. Open your Hubarr instance in Chrome/Firefox
+#   2. DevTools → Application → Cookies → find "hubarr_session"
+#   3. Copy the Value column and paste it here
+#   Only needed on first run (or when your session expires).
+SESSION_COOKIE=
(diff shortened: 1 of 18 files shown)


Migz93:
Codex-generated follow-up review: I re-checked the latest changes against the earlier review feedback and I agree with the fixes.

- `tests/playwright/posters.spec.ts` now scopes the selector to actual Plex poster images, so unrelated avatar/image failures no longer cause misleading regressions.
- The poster check now skips lazy images that have not loaded yet, which avoids false failures for offscreen items while still failing genuinely broken loaded images.
- `tests/playwright/auth.setup.ts` now detects when the saved storage state was created for a different host than the current `BASE_URL`, so it correctly re-authenticates instead of reusing cookies scoped to the wrong domain.

I also ran `npm run check` locally and it passed.

I did not run the Playwright suite itself here because that still depends on a live Hubarr instance plus a valid `.env.playwright` session cookie.

PR #31, opened 2026-04-09
feat: apply isolation filters to Plex Home managed users

Migz93:
## Summary

- Plex Home managed users (sub-accounts with no independent Plex.tv account) were being silently skipped during isolation filter sync because the `sharing_settings` API requires an email address to identify the user — managed accounts have none
- Investigation revealed the same `sharing_settings` endpoint also accepts `invitedId` (numeric user ID) as an alternative identifier, which works for managed accounts
- Managed users with a restriction profile (e.g. "Younger Kid") are intentionally skipped — Plex prevents label filter changes on those accounts anyway
- A new read-only "Managed Home Users" section is added to the Users page so they're visible, with a note explaining watchlists aren't available for managed accounts

## Changes

- **`plex.ts`**: Added `getHomeOnlyManagedUsers()` which fetches `plex.tv/api/users`, filters for `home="1"` accounts with no username/email on this server, and detects restriction profiles via `contentRating=` in filter strings. Updated `syncIsolationFilters()` and `clearHubarrIsolationFilters()` to process these users using `invitedId` in the payload.
- **`services.ts`**: Added `getManagedUsers()` method.
- **`app.ts`**: Added `GET /api/users/managed` endpoint (registered before `/:id` to avoid route conflict).
- **`shared/types.ts`**: Added `ManagedUserRecord` interface.
- **`Users.tsx`**: Fetches managed users on load and renders a collapsible read-only section at the bottom of the Users page.

## Test plan

- [ ] Run a publish pass and confirm managed users without a restriction profile now have `label!=hubarr:*` exclusion filters applied in Plex
- [ ] Confirm managed users with a restriction profile (e.g. Hugo) are skipped and their existing filters are untouched
- [ ] Check the Users page shows a "Managed Home Users" section listing managed accounts as read-only
- [ ] Confirm "Refresh Users" still works and the managed section updates

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ Hubarr uses a few background jobs:
 
 - **Watchlist RSS Sync** watches Plex RSS feeds for quick watchlist changes and stores new items fast
 - **Watchlist GraphQL Sync** regularly performs a fuller watchlist reconciliation to catch anything RSS missed
-- **Plex library scans** help Hubarr notice when something from a watchlist has now actually appeared in your Plex libraries
+- **Plex Library Scans** help Hubarr notice when something from a watchlist has now actually appeared in your Plex libraries
 - **Collection Sync** is the job that updates the Plex collections and hub rows
 
 Together, that means Hubarr can react quickly when watchlists change, while still having a slower safety net that keeps everything accurate over time.
@@ -101,6 +101,14 @@ Hubarr is configured through its web UI after first run. The two things you may 
 5. Run a sync, or wait for the scheduled jobs to start working
 
 ## Important Limitations
+
+### Plex Home Managed Users
+
+Plex Home managed users (sub-accounts with no independent Plex account) have two limitations:
+
+- **Watchlists cannot be synced.** Managed users have no presence in the Plex community API, so Hubarr cannot read their watchlist. They will appear in the Users page as read-only and cannot be enabled.
+- **Label exclusions are applied automatically.** Even though their watchlist can't be synced, Hubarr will still apply label exclusion filters to managed users so that other users' watchlist collections don't appear for them. This happens as part of the normal collection sync.
+- **Restriction Profiles block label exclusions.** If a managed user has a Plex restriction profile set (e.g. Younger Kid, Older Kid, Teen), Plex prevents label-based filter changes for that account. Hubarr skips those users entirely — the restriction profile itself is usually sufficient exclusion anyway.
 
 ### Plex Friends Home Privacy Caveat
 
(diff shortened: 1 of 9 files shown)


Migz93:
Codex reviewed this PR and found two issues worth addressing before merge:

1. `clearHubarrIsolationFilters()` can still try to update restricted managed users, even though `syncIsolationFilters()` explicitly skips them because Plex rejects label filter changes for accounts with restriction profiles. If one of those users ever has a `hubarr:*` filter present, reset can fail and abort the wider reset flow.

Relevant code:
- `src/server/integrations/plex.ts` in the managed-user branch of `syncIsolationFilters()`
- `src/server/integrations/plex.ts` in the managed-user branch of `clearHubarrIsolationFilters()`
- `src/server/services.ts` in `resetCollections()`

2. The Users page now depends on `/api/users/managed` inside a single `Promise.all(...)` alongside the existing `/api/users` and `/api/settings` requests. That means a transient Plex.tv failure in the new managed-users endpoint can break the whole Users page, including the existing editable user list, even though the new section is read-only and supplemental.

Relevant code:
- `src/client/pages/Users.tsx` in `load()`
- `src/server/app.ts` in `GET /api/users/managed`

I also ran local validation during the review: `npm run check` and `npm run build` both passed.

Migz93:
Codex reviewed the latest follow-up commit on this PR. The earlier issues I flagged do look addressed: the clear path now skips restriction-profile managed users, and the Users page no longer calls Plex.tv on every page load. I do still see two remaining issues:

1. Stale managed users can persist forever in the DB cache

`upsertManagedUsers()` only inserts/updates rows and never removes rows that are no longer returned by Plex Home. Because `/api/users/managed` now reads directly from that cached table, a managed user removed from Plex Home can keep appearing in the UI indefinitely until the database is manually cleaned.

Relevant code:
- `src/server/db/users.ts` in `upsertManagedUsers()`
- `src/server/services.ts` in `discoverUsers()` / `getManagedUsers()`

2. Refresh Users is still all-or-nothing on the managed-user fetch

`discoverUsers()` now does `Promise.all([plex.discoverUsers(), plex.fetchManagedUsers()])`. That means a transient failure in the managed-user Plex.tv call will cause the whole refresh to fail, so regular Plex friend discovery will not update either. This is an improvement over breaking the Users page on load, but it still makes the refresh path more fragile than it used to be.

Relevant code:
- `src/server/services.ts` in `discoverUsers()`

Local validation during this review:
- `npm run check`
- `npm run build`

Both passed.

Migz93:
Codex reviewed the latest follow-up commit and I’m not seeing any further issues in the changes.

The two remaining review points appear addressed:

- stale managed users are now pruned from the cache during refresh
- managed-user fetch failures no longer block normal user discovery

I also re-ran local validation during this pass:
- `npm run check`
- `npm run build`

Both passed.

Residual risk is mainly around live Plex behavior, since the managed-user refresh/cache path and restriction-profile handling are still relying on manual verification rather than automated coverage, but I don’t have any new code review findings on the latest commit.

PR #33, opened 2026-04-09
feat: replace image proxy with local disk cache

### LINKED ISSUE #28: feat: replace image proxy with local image cache

Migz93:
Closes #28

## Summary

- Removes the `/api/plex/image` and `/api/avatar` proxy endpoints entirely — no transition behaviour, clean cutover
- Images are downloaded and resized at sync time via a new `ImageCacheService` (uses `sharp`), stored under `/config/image-cache/`, and served as authenticated static files from `/images/`
- DB migration v3 adds nullable `cached_thumb` / `cached_avatar_url` columns; all queries `COALESCE(cached, raw)` so rows without a cached image still return the raw URL as a fallback during the first sync
- `getPlexImageSrc` now only returns `/images/` paths — external URLs are treated as null, showing the generic fallback SVG
- Self-user avatar is refreshed on every full sync (previously only on Plex settings save), and the sidebar admin icon reads `cached_avatar_url` from the DB rather than the raw URL stored in the session record
- Generic `poster-fallback.svg` displayed on all poster `` elements until the cache is populated
- "Clear Image Cache" button in Settings → General wipes disk files and resets DB paths
- Playwright tests: `/images/` auth guard, dashboard poster loading, watchlist poster loading, user avatar loading

## Test plan

- [ ] Run a **Watchlist GraphQL Sync** — verify `/config/image-cache/` is populated with `.jpg` files and posters appear on Dashboard and Watchlists
- [ ] Check Users page — avatars load from `/images/`
- [ ] Check sidebar bottom-left admin icon loads correctly
- [ ] Press **Clear Image Cache** in Settings → General — files removed, count shown, fallback SVG displayed
- [ ] Run sync again — images re-populate
- [ ] Run `npm run test:e2e` — all 43 tests pass

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -75,11 +75,13 @@ Read-only. Safe to run against a live instance.
 
 ---
 
-### `tests/playwright/posters.spec.ts` — Poster image loading
+### `tests/playwright/posters.spec.ts` — Image cache
 
-Requires a working Plex connection. Poster images are served through the `/api/plex/image` proxy — failures here typically mean the Plex connection is broken or a specific poster URL is bad. Items with no poster URL render a fallback icon rather than an `<img>`, so they are naturally excluded from these checks.
+Tests that cached images are served correctly from `/images/`, that the route is protected, and that user avatars load. Images are cached at sync time — tests log and skip gracefully if no images have been cached yet (run a full sync first).
 
 | Test | What it checks |
 |---|---|
-| Dashboard recently added posters all load | Waits for the dashboard to finish loading, then checks every `img.object-cover` element has loaded successfully (`complete && naturalWidth > 0`). Logs and skips gracefully if there are no poster images yet. |
-| Watchlists page 1 posters all load | Same check on the first page of the Watchlists grid. |
+| `/images/` route requires authentication | Opens a fresh context with no session and requests `/images/test.jpg` — expects a 401 or redirect to login |
+| Dashboard recently added posters all load | Waits for the dashboard to finish loading, then checks every `img.object-cover[src*='/images/']` has loaded successfully (`complete && naturalWidth > 0`) |
+| Watchlists page 1 posters all load | Same check on the first page of the Watchlists grid |
+| Users page avatar images load from /images/ or show fallback | Checks every `img[src*='/images/']` on the Users page has loaded successfully |
--- a/docs/image-caching.md
+++ b/docs/image-caching.md
@@ -1,0 +1,298 @@
+# Image Caching
+
+This document describes the image caching system in Hubarr — how posters and
+avatars are fetched, stored, served, and refreshed.
+
+## Overview
+
+Hubarr caches all images locally at sync time. The browser never talks to Plex
+or any external CDN for images. Every poster and avatar served to the UI comes
+from a file on disk under `/config/image-cache/`, served through the
+authenticated `/images/` static route.
+
+The system has three operating modes for any given image:
+
+| Situation | Behaviour |
+|---|---|
+| No cached entry exists | Fetch from upstream inline, write to disk, record metadata |
+| Cached entry exists, file is fresh | Return local path immediately, do nothing |
+| Cached entry exists, file is stale | Return existing local path immediately, start a background refresh |
+
+A failed background refresh never removes the existing file. The stale image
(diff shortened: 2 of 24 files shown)


Migz93:
CODEX review:

I did a deep pass over the image-cache changes and found one blocking issue plus one lower-severity mismatch to clean up.

1. High: external-image caching reintroduces SSRF via redirect following. In both poster and avatar fetch paths, the code sanitizes only the initial URL, then calls `fetch(safe)` with the default redirect behavior, which will follow upstream `3xx` responses automatically. That means a seemingly safe HTTPS URL can bounce the server to an internal or otherwise disallowed target without any second validation step. This is especially relevant because the old `/api/avatar` proxy explicitly validated every redirect hop; the new cache path no longer does. See `src/server/image-cache.ts:484`, `src/server/image-cache.ts:496`, `src/server/image-cache.ts:519`, and `src/server/plex-image-utils.ts:20`. I’d treat this as a merge blocker and either set `redirect: "manual"` and re-sanitize `Location` on each hop, or use `redirect: "error"`.

2. Low: the user-facing refresh policy is now inconsistent with the implementation. Settings says cached images refresh “after 7 days,” but the code actually uses upstream `Cache-Control: max-age` when present and otherwise falls back to 30 days. That will make troubleshooting confusing because the UI and runtime behavior disagree. See `src/client/pages/Settings.tsx:246` and `src/server/image-cache.ts:635`.

Aside from that, I did not find other blocking correctness issues in the changed paths. This review was done by CODEX.

Migz93:
CODEX follow-up review:

One new issue from the latest hardening commit:

1. Medium: the redirect-hardening fix may now block legitimate redirects for Plex-hosted images. `fetchFollowingRedirects()` uses `sanitizeAvatarUrl()` for every redirect hop, but that sanitizer only allows `https:` and rejects private/loopback hosts. That makes sense for external avatar/CDN URLs, but it is also now used for `plex-path` fetches started from the user's own Plex server URL. If Plex returns a relative redirect or redirects to another local Plex host, that hop will now be rejected even though it is part of the trusted Plex flow. See `src/server/image-cache.ts:462`, `src/server/image-cache.ts:542`, and `src/server/plex-image-utils.ts:20`. I'd want either a Plex-specific redirect validator or a clear confirmation that these endpoints never redirect in real deployments.

This review was done by CODEX.

Migz93:
CODEX follow-up review:

I re-reviewed the latest commit, and I do not see a blocker from this change. The trust split now looks correct: trusted Plex-relative image fetches use native redirect following again, while untrusted external poster/avatar URLs still go through the manual hop-by-hop validator. That addresses the regression risk I called out previously without reopening the original external redirect issue.

Residual risk is mostly policy rather than implementation: trusting Plex redirects means the configured Plex server is explicitly inside the trust boundary. If that assumption is acceptable for Hubarr, this looks reasonable to me.

Local `npm run check` passed during review. This review was done by CODEX.

PR #39, opened 2026-04-10
feat: redesign users page with responsive card grid layout

### LINKED ISSUE #38: enhancement: redesign users page layout to make better use of space
The current users page shows each user in a full-width vertical list, which leaves a lot of unused space and makes the page feel less efficient than it could be.

It would be worth redesigning this page to make better use of the available width, especially when viewing multiple users at once.

One possible direction would be to display users as cards or tiles in a grid rather than as full-width rows.

Goals:
- Reduce wasted horizontal space on larger screens
- Make it easier to scan multiple users at once
- Explore a card or square-based layout instead of a full-width vertical list
- Keep the layout clear and usable on smaller screens as well

This is primarily a UX/UI improvement, so the investigation should focus on finding a layout that feels more compact and readable without making the page feel cluttered.

Migz93:
## Summary

- Replaces full-width row list with a responsive card grid (2 cols on mobile → 6 cols on large screens), matching the dashboard poster grid pattern
- Each card shows a large circular avatar and display name; the **You** badge sits inline next to the name for the logged-in user
- Managed users get a clickable **Managed User** badge (click to see: "Watchlists are not available for managed users") and, where applicable, a clickable **Restriction Profile** badge (click to see: "Label exclusion cannot be applied to user with Restriction Profile") — clicking outside the popup dismisses it
- Checkbox selection, bulk enable/disable, **Sync Watchlist** button, edit pencil, collapsible Disabled and Managed Users sections, and the edit modal are all preserved

Closes #38

## Test plan

- [ ] Active users render as a grid; scales from 2 cols on mobile to 6 on wide screens
- [ ] **You** badge appears inline next to the logged-in user's name
- [ ] Sync Watchlist button present and functional on active user cards; absent on disabled cards
- [ ] Edit pencil opens the existing edit modal
- [ ] Checkbox selection + bulk enable/disable works
- [ ] Managed Users section (collapsed by default) renders cards with Managed User badge
- [ ] Clicking Managed User badge shows popup message; clicking outside dismisses it
- [ ] Restriction Profile badge appears only where applicable; popup message is correct

### DIFF
--- a/src/client/pages/Users.tsx
+++ b/src/client/pages/Users.tsx
@@ -149,16 +149,16 @@ export default function Users() {
         </div>
       </div>
 
-      {disabledUsers.length > 0 && (
-        <div className="mb-6">
-          <button
-            onClick={() => setDisabledOpen((open) => !open)}
-            className="flex items-center gap-2 text-sm font-medium text-on-surface-variant mb-3"
-          >
-            {disabledOpen ? <ChevronDown size={16} /> : <ChevronRight size={16} />}
-            Disabled ({disabledUsers.length})
-          </button>
-          {disabledOpen && (
+      <div className="mb-6">
+        <button
+          onClick={() => setDisabledOpen((open) => !open)}
+          className="flex items-center gap-2 text-sm font-medium text-on-surface-variant mb-3"
+        >
+          {disabledOpen ? <ChevronDown size={16} /> : <ChevronRight size={16} />}
+          Disabled ({disabledUsers.length})
+        </button>
+        {disabledOpen && (
+          disabledUsers.length > 0 ? (
             <div className="grid grid-cols-2 sm:grid-cols-3 md:grid-cols-4 lg:grid-cols-5 xl:grid-cols-6 gap-3">
               {disabledUsers.map((user) => (
                 <UserCard
@@ -172,9 +172,11 @@ export default function Users() {
                 />
               ))}
             </div>
-          )}
-        </div>
-      )}
+          ) : (
+            <p className="text-sm text-on-surface-variant">No disabled users</p>
+          )
+        )}
+      </div>
 
       {managedUsers.length > 0 && (
         <div>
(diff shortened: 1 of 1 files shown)


Migz93:
CODEX review:

I found two things I’d recommend changing before merge.

1. The new centered card header looks like it can lose reliable truncation for long user names. In `src/client/pages/Users.tsx`, the display name now sits in a flex row with the inline `You` badge, but without the old shrinking constraints, so long names may overflow the card instead of truncating cleanly on narrower breakpoints.

2. I’d keep the Disabled section visible even when there are zero disabled users. Hiding it entirely changes the page structure and also conflicts with the current Users page test expectations. A better UX here would be to always show `Disabled (0)` and, when expanded with no entries, render a small muted empty state like `No disabled users`.

Those are the only issues I found in the PR; the rest of the redesign looks good from my review.

Migz93:
CODEX follow-up review:

The Disabled section change looks good now, but I still think the truncation fix needs one more pass.

In `src/client/pages/Users.tsx`, adding `min-w-0` to the flex container helps, but the actual truncating flex item is still the display-name `<p>`. In this kind of flex row, that element can still keep its intrinsic min width and refuse to shrink, so long names may continue to overflow next to the inline `You` badge.

I’d suggest putting the shrinking constraint on the text element itself, for example `min-w-0 flex-1 truncate`, or wrapping it in a `min-w-0 flex-1` container and keeping `truncate` there.

Apart from that, the previous review points look addressed.

Posted from CODEX.

Migz93:
CODEX review update:

I’ve re-reviewed the latest changes and I’m happy with this PR now.

The previous review points have been addressed:
- the Disabled section remains visible and shows a sensible empty state when there are no disabled users
- long usernames now truncate correctly in the card layout next to the inline `You` badge

You also validated the updated UI against the live container and confirmed Playwright was fine, including a manual check with an intentionally long username.

From CODEX’s side, this is approved.

PR #40, opened 2026-04-10
feat: introduce true release-date sorting for Plex collections

### LINKED ISSUE #37: bug: review "date new to old" collection ordering
The collection ordering option for `date new to old` may not be using the correct date field when sorting items.

In at least one case, Hubarr is showing the new Demon Slayer movie as more recent than Zootopia 2, which appears to be incorrect.

This suggests the sort may be using the wrong date source, or that the intended date field for ordering is unclear.

Things to check:
- Which date field is currently used for `date new to old`
- Whether the sort should use release date, digital release date, theatrical date, watchlist added date, or another value
- Whether missing or partial dates are causing incorrect ordering
- Whether the UI label matches the actual sort behavior

Expected behaviour:
Items should be ordered consistently according to the intended "newest first" date, and the sort should not place older titles ahead of newer ones when the underlying dates do not support that order.

Migz93:
Fixes #37

## Summary

- Replaces year-based collection ordering with full date-precision sorting using `originallyAvailableAt` from Plex discover metadata
- Adds `releaseDate: string | null` to `WatchlistItem`, persisted in `raw_payload` (no schema migration needed)
- Fixes a longstanding inconsistency where ascending and descending sort used completely different code paths (Hubarr-managed vs Plex-native)
- Fixes RSS-ingested items bypassing the discover metadata lookup when they already had `thumb` + `year`, meaning they never gained a real release date

## What changed

**New field — `WatchlistItem.releaseDate`**
ISO date string (`YYYY-MM-DD`) from Plex discover `originallyAvailableAt`. Falls back to synthesised `YYYY-01-01` from `year` when only a year is available. Kept separate from `addedAt` (watchlist entry time) which is not a media release date.

**Enrichment shortcut fixed**
`enrichWatchlistMetadata` previously skipped the discover lookup when `thumb && year` were both set. The new rule is `thumb && releaseDate` — so RSS items with a poster and year still go through discover to get a proper date.

**Both sort directions now Hubarr-managed**
A new `compareByReleaseDate` helper sorts by date → title → plexItemId (nulls last in both directions). Both `date-desc` and `date-asc` set Plex `collectionSort=2` (custom) and explicitly reorder items, rather than `date-asc` delegating to Plex native sort.

**Renamed sort values with backward compat**
`year-desc` → `date-desc`, `year-asc` → `date-asc`. Existing stored settings are normalised on every read in `getAppSettings` — no migration, no user action needed.

**UI labels updated**
"Release Year (New to Old/Old to New)" → "Release Date (New to Old/Old to New)"

## Testing

- Playwright suite passes
- `npm run check` and `npm run build` both clean
- Tested locally against rebuilt container (port 9301); collection syncs running correctly in logs

### DIFF
--- a/src/client/components/CollectionsConfigForm.tsx
+++ b/src/client/components/CollectionsConfigForm.tsx
@@ -105,8 +105,8 @@ export default function CollectionsConfigForm({
               setForm((current) => ({ ...current, collectionSortOrder: value as CollectionSortOrder }))
             }
           >
-            <option value="year-desc">Release Year (New to Old)</option>
-            <option value="year-asc">Release Year (Old to New)</option>
+            <option value="date-desc">Release Date (New to Old)</option>
+            <option value="date-asc">Release Date (Old to New)</option>
             <option value="title">Title (A–Z)</option>
           </SelectInput>
         </Field>
--- a/src/server/app.ts
+++ b/src/server/app.ts
@@ -583,9 +583,12 @@ export function createApp(config: RuntimeConfig, scheduler?: JobScheduler) {
         patch.collectionNamePattern = body.collections.collectionNamePattern;
       }
       if (body.collections.collectionSortOrder !== undefined) {
-        const valid = ["year-desc", "year-asc", "title"];
-        if (valid.includes(body.collections.collectionSortOrder)) {
-          patch.collectionSortOrder = body.collections.collectionSortOrder as "year-desc" | "year-asc" | "title";
+        // Accept new date-* values and normalize legacy year-* values on ingest.
+        const legacyMap: Record<string, string> = { "year-desc": "date-desc", "year-asc": "date-asc" };
+        const normalized = legacyMap[body.collections.collectionSortOrder] ?? body.collections.collectionSortOrder;
+        const valid = ["date-desc", "date-asc", "title"];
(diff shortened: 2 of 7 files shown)


Migz93:
Codex review note:

This PR looks good to merge for the `#37` fix.

What looks solid:
- The sort semantics are now correct: `releaseDate` is separated from `addedAt`, so watchlist/feed timing is no longer being used as a proxy for media release date.
- RSS items no longer permanently bypass enrichment when they already have a thumb + year.
- Both sort directions now use the same Hubarr-managed custom ordering path instead of one direction being Hubarr-managed and the other relying on Plex native sorting.
- Backward compatibility for stored `year-*` sort settings appears to be handled cleanly.

Agreed rollout caveat:
- Immediately after deploy, some existing cached items may sort imperfectly until they’ve been refreshed by the first RSS or GraphQL sync, because older cached rows won’t have `releaseDate` populated yet.
- This was reviewed and considered acceptable for this change, since it should self-correct on the first sync cycle rather than needing manual intervention.

Verification status recorded for this review:
- Playwright had no issues.
- `npm run check` had no issues.
- `npm run build` had no issues.

From this Codex review pass, this is fine to merge.

PR #57, opened 2026-04-11
feat: preserve original watchlist addedAt dates via activity cache and RSS

### LINKED ISSUE #36: enhancement: investigate preserving original watchlist added dates
When Hubarr is set up on a fresh instance, items imported from a person's watchlist are currently assigned the current date as their "added" date.

This means the timeline in Hubarr does not reflect when those items were actually added to the source watchlist, and everything appears as if it was added on the day Hubarr was first configured.

We should investigate whether the upstream watchlist source exposes the original date each item was added to the watchlist, and if so, whether Hubarr can use that value when importing items.

Goals:
- Find out whether the watchlist provider exposes an original "date added" field
- Confirm whether that value is available through the API/library we use
- Assess whether Hubarr can store and display that original date instead of using the import date
- Identify any limitations, such as missing dates for older entries or inconsistent support across sources

If supported, a fresh Hubarr setup should preserve the real watchlist added dates rather than assigning everything today's date.

Migz93:
## Summary

- Introduces a `watchlist_activity_cache` table populated from the Plex Community `activityFeed` GraphQL query, storing the real date each item was watchlisted per user
- Uses a sentinel value (`2001-01-01`) for unknown dates rather than today's date, so real dates can be resolved later and the UI can show "Unknown" meaningfully
- Ad-hoc sync buttons (dashboard and per-user) now run a full pipeline: refresh activity cache → RSS date snapshot → GraphQL sync → immediate collection publish
- RSS and GraphQL syncs trigger an immediate collection publish when changes are detected, rather than waiting for the next scheduled job
- Fixes RSS deduplication so two friends watchlisting the same item are both detected and processed
- Adds History page section surfacing items whose `addedAt` date could not be resolved
- Replaces `plans/` docs with proper `docs/` reference documentation covering the full watchlist pipeline

### Server
- `migrations.ts` — v5: `watchlist_activity_cache` and `job_run_state` tables; v6: `activityCacheFetchIntervalMinutes` setting
- `plex.ts` — `fetchWatchlistActivityFeed()`, `fetchSelfPlexUuid()`, `fetchSelfWatchlist()`
- `services.ts` — `syncActivityCache()`, `buildAllRssDateMaps()`, extended `syncUser()` with three-step date resolution, `runFullSync()` and `runUserSync()` now chain activity cache + RSS + publish, `pollRss()` triggers publish on new items
- `rss-cache.ts` — author included in stable key so two friends watchlisting the same item are distinct entries
- `db/` — `getActivityCacheDate()`, `upsertActivityCacheEntries()`, `getJobRunState()`, `saveJobRunState()`
- `app.ts` — activity cache and job state API endpoints

### Client
- Settings: Activity Cache Fetch interval configurable under Jobs
- History: Watchlisted At Date Unresolved collapsible section
- Watchlists / Dashboard / WatchlistItemModal: sentinel-aware `addedAt` display ("Unknown" instead of 01/01/2001)

### Docs
- `docs/watchlist.md` — full end-to-end watchlist pipeline reference
- `docs/sync.md` — sync architecture, job model, settings, startup sequence
- `docs/architecture.md` — high-level architecture overview
- `plans/` directory removed

## Test plan

- [ ] First setup: dashboard sync button fetches activity cache, resolves dates, publishes collections in one shot
- [ ] Per-user sync button does the same scoped to that user
- [ ] Items with no resolvable date show "Unknown" in UI and appear in History under "Watchlisted At Date Unresolved"
- [ ] Items with resolved dates show correct date and do not appear in unresolved section
- [ ] RSS new item detection works for two friends watchlisting the same item (both processed)
- [ ] RSS processing triggers immediate collection publish
- [ ] Scheduled GraphQL sync triggers immediate collection publish when items change
- [ ] Activity Cache Fetch interval is configurable in Settings > Jobs

Closes #36

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -4,9 +4,13 @@
 
 Hubarr is a single-purpose Plex companion app.
 
-It tracks the owner's Plex watchlist and selected friends' watchlists, stores that watchlist state locally, matches items against Plex libraries, and keeps per-user Plex collections and hub rows up to date.
+It tracks the owner's Plex watchlist and selected friends' watchlists, stores
+that watchlist state locally, matches items against Plex libraries, and keeps
+per-user Plex collections and hub rows up to date.
 
 Hubarr does not request or download media.
+
+---
 
 ## Core Model
 
@@ -15,6 +19,8 @@ Hubarr does not request or download media.
 - Each enabled user gets one movie collection and one TV collection
 - Both collections share the same visible name for that user
 - Hubarr is a single-owner app: the first Plex account to onboard becomes the only admin
+
+---
 
 ## Deployment Model
 
@@ -29,12 +35,13 @@ Hubarr runs as a single self-hosted container:
 
 Persistent data is stored in `/config`.
 
+---
+
 ## Database Migrations
 
 Hubarr uses SQLite `PRAGMA user_version` for schema migrations.
 
 - Versioned migrations live in `src/server/db/migrations.ts`
-- The current baseline schema is migration `1`
 - `runMigrations(db)` runs on startup, applies any migration whose version is higher than the current `user_version`, and advances `user_version` after each successful migration
 - Each migration runs inside a transaction so a failure should leave the database unchanged
 
@@ -44,6 +51,8 @@ When changing the schema in the future:
 2. Write the schema change in that migration's `up(db)` function
 3. Do not edit older migrations that may already have shipped
 4. Keep default-setting seeding separate from schema migrations
+
+---
 
 ## Auth And Setup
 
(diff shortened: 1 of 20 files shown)


Migz93:
Addressed all three review findings — pushed in e5966ae.

**CodeQL — unused import**
Removed `formatRelativeTime` from the `WatchlistItemModal.tsx` import; it was imported but never referenced in the file.

---

**Codex P1 — cursor not advanced on failed fetch**
In `syncActivityCache()`, the catch block was saving `lastRunAt: new Date().toISOString()`, which advanced the incremental fetch cursor even when the fetch failed. The next run would then start *after* the failed window, permanently skipping any activity events from that period and leaving `addedAt` unresolved for affected items.

Fixed by saving `lastRunAt: since` in the catch block — the cursor stays at the pre-failure value so the next run retries from the same point.

---

**Codex P2 — errors swallowed by scheduler wrapper**
The `activity-cache-fetch` task in `index.ts` wrapped `syncActivityCache()` in a try/catch and swallowed the error. The scheduler's `execute()` saw the task complete without throwing, so it persisted `lastRunStatus: "success"`, overwriting the `error` state the service had written. The Jobs UI always showed success regardless of what actually happened.

Fixed in two parts:
1. Removed the catch-and-swallow in `index.ts` so errors propagate to the scheduler.
2. Fixed `JobScheduler.execute()` to not advance `lastRunAt` on error — without this the scheduler would have immediately re-advanced the cursor the P1 fix just preserved.

The Jobs UI will now correctly show `error` status for failed runs, and `lastRunAt` reflects the last *successful* run (keeping it valid as a fetch cursor).

PR #59, opened 2026-04-11
feat: watchlisted date sorting and per-user collection order overrides

Migz93:
## Summary

- Adds two new collection sort modes — **Watchlisted Date (New to Old)** and **Watchlisted Date (Old to New)** — sorting each user's Plex collection by when they added items to their watchlist, using the `addedAt` timestamp already tracked by the activity cache
- Adds a **per-user collection ordering override** (`collection_sort_order_override` column, DB migration v6); null means inherit the global setting, matching the existing pattern for `visibilityOverride`
- Items with no resolved watchlist date (added before the activity cache existed) sort to the end and emit a `warn` log listing their titles so the cause is visible

## Changes

- `src/shared/types.ts` — extended `CollectionSortOrder`, added `collectionSortOrderOverride` to `UserRecord`
- `src/server/db/migrations.ts` — migration v6 adds `collection_sort_order_override TEXT` to `users`
- `src/server/db/users.ts` — `listUsers()` and `updateUser()` updated for the new column
- `src/server/app.ts` — whitelist updated in `PATCH /api/settings`; validation added in `PATCH /api/users/:id`
- `src/server/services.ts` — new `compareByWatchlistDate()` using `Date.getTime()` for format-robust comparison; `publishUserCollections()` resolves effective sort order from per-user override and logs when active
- `src/server/integrations/plex.ts` — `plexSort` map extended for both new values (→ `collectionSort=2`)
- `src/client/components/CollectionsConfigForm.tsx` — two new options in the global ordering dropdown
- `src/client/pages/Users.tsx` — sort override section in the user edit modal; hint shows human-readable label for the global default
- `tests/playwright/settings.spec.ts` / `users.spec.ts` — structural smoke tests for the new UI elements
- `TESTING.md` — test table updated

## Test plan

- [ ] Settings → Collections tab: verify "Watchlisted Date (New to Old)" and "Watchlisted Date (Old to New)" appear in the ordering dropdown
- [ ] Users page → edit a user: verify "Collection Ordering" section appears with dropdown and "Restore to global default" reset; hint shows the global sort label in plain English (e.g. "Release Date (New to Old)")
- [ ] Set global sort to a watchlist date mode and trigger a collection publish; verify Plex ordering reflects each user's `addedAt`
- [ ] Set a per-user override and verify that user's collection uses the override while others use the global setting
- [ ] Items with no watchlist date appear at the end and a `warn` log entry lists them by title
- [ ] Run `npm run test:e2e` against a live instance

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -85,6 +85,34 @@ Read-only. Safe to run against a live instance.
 
 ---
 
+### `tests/playwright/settings.spec.ts` — Settings tabs
+
+Read-only. Safe to run against a live instance.
+
+| Test | What it checks |
+|---|---|
+| All six tabs are visible | Navigates to `/settings`, asserts all six tab buttons (General, Plex, Collections, Logs, Jobs, About) are rendered |
+| Clicking a tab updates the URL | Clicks Plex, Jobs, and About tabs in turn; asserts the URL gains the expected `?tab=` parameter |
+| General tab shows Startup Sync toggle and History Retention field | Navigates to `/settings?tab=general`, waits for load, asserts both setting labels are visible |
+| Jobs tab shows the jobs table | Navigates to `/settings?tab=jobs`, asserts the "Job Name" column header is visible |
+| About tab shows version and support info | Navigates to `/settings?tab=about`, asserts "About Hubarr" and "Version" headings are visible |
+| Collections tab shows watchlisted date sort options | Navigates to `/settings?tab=collections`, asserts the ordering `<select>` contains the "Watchlisted Date (New to Old)" and "Watchlisted Date (Old to New)" options |
+
+---
+
+### `tests/playwright/users.spec.ts` — Users page structure
+
+Read-only. Safe to run against a live instance.
+
+| Test | What it checks |
+|---|---|
+| Active users section heading is visible | Asserts the "Active (N)" heading renders |
+| Disabled users accordion toggle is visible | Asserts the "Disabled (N)" toggle button renders |
+| Refresh Users button is present | Asserts the Refresh Users button renders |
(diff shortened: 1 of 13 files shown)


Migz93 (commented):
Two findings from review:

1. `src/server/db/index.ts` still exposes `updateUser()` with the old patch type, so it does not include `collectionSortOrderOverride`, even though `src/server/db/users.ts` now supports it and the route passes it through. The feature works today because the route is using a loose `Record<string, unknown>`, but any typed caller of the DB facade will be blocked from using the new override or could lose it during a later refactor.

2. `src/server/services.ts` introduces a second hardcoded unknown-date sentinel instead of reusing `WATCHLIST_DATE_UNKNOWN_SENTINEL` from `src/server/integrations/plex.ts`. The new watchlist-date sort and warning path now depend on those two literals staying in sync. If the shared sentinel ever changes, unresolved items would stop being pushed to the end and would stop generating the warning this PR promises.

`npm run check` and `npm run build` both passed locally. Residual risk: the new Playwright coverage is structural smoke coverage only, so it does not verify persistence of the override or end-to-end Plex ordering behavior.

Migz93:
Follow-up review: the two findings from my earlier review look resolved in the latest commit.

- `src/server/db/index.ts` now includes `collectionSortOrderOverride` in the `updateUser()` facade type, so the DB wrapper matches the repository implementation.
- `src/server/services.ts` now reuses `WATCHLIST_DATE_UNKNOWN_SENTINEL` instead of maintaining a duplicate local sentinel.

I reran `npm run check` and `npm run build` locally and both passed. I don’t have any further review findings on this PR.

- Codex

PR #62, opened 2026-04-12
feat: add polling-based live UI refresh

### LINKED ISSUE #46: enhancement: add live UI updates across the app
A number of pages in Hubarr currently require a manual browser refresh before newly changed data appears in the UI.

This makes the app feel less active and responsive than it could be, especially when Hubarr is already processing new information correctly in the background.

Examples observed:
- If a new item is added to a watchlist and Hubarr picks it up, it does not appear on the dashboard or watchlists page until the page is refreshed
- On the homepage, a new entry in `Recent Syncs` does not appear until refresh, and the status does not update from running to success without another refresh
- If an item changes from missing from library to in library, the UI does not reflect it until refresh
- On the history page, new activity is not shown without refreshing
- On `Settings > Jobs`, a job starting or finishing is not reflected until refresh
- On the users page, refreshed user data may also require a manual page reload to appear (this needs confirming)

This would be worth reviewing as a broader UX issue rather than page-by-page in isolation.

Goals:
- Identify the parts of the UI that rely too heavily on manual refresh
- Improve how the frontend reacts to background changes and completed jobs
- Make key views feel more live and up to date
- Ensure status changes are reflected automatically where it matters most
- Review whether polling, invalidation, or a more real-time approach would be the best fit

The aim is for Hubarr to feel more active and responsive, so that new items, sync activity, status changes, and job state updates appear automatically without requiring the user to refresh the page.

Migz93:
## Summary
- add a shared polling-based live refresh helper for client pages
- live-refresh dashboard, watchlists, history, jobs, and users
- move Refresh Users onto a scheduled background job and normalize jobs running state
- update docs and add Playwright coverage for the live-refresh flows

## Verification
- npm run check
- npx playwright test tests/playwright/users.spec.ts
- npx playwright test tests/playwright/live-refresh.spec.ts

Closes #46

🤖 Generated with Codex

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -123,3 +123,16 @@ Tests that cached images are served correctly from `/images/`, that the route is
 | Dashboard recently added posters all load | Waits for the dashboard to finish loading, then checks every `img.object-cover[src*='/images/']` has loaded successfully (`complete && naturalWidth > 0`) |
 | Watchlists page 1 posters all load | Same check on the first page of the Watchlists grid |
 | Users page avatar images load from /images/ or show fallback | Checks every `img[src*='/images/']` on the Users page has loaded successfully |
+
+---
+
+### `tests/playwright/live-refresh.spec.ts` — Live refresh behavior
+
+These tests trigger real background work and verify that the open page updates
+without a browser reload. They are not read-only.
+
+| Test | What it checks |
+|---|---|
+| Dashboard recent syncs updates after a background collection sync starts and finishes | Opens `/dashboard`, triggers the collection-sync job through the API, and verifies the `Recent Syncs` panel first shows a running Publish entry and then updates to the finished summary without a reload |
+| History shows a new collection sync row move from running to its terminal status without reload | Opens `/history`, triggers the collection-sync job through the API, and verifies the newest row appears as running and then changes to its final status/summary automatically |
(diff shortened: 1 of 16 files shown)


Migz93:
Two minor items worth noting from the review:

**1. Unnecessary dep in the cleanup effect (`useLiveRefresh.ts`)**

`scheduleNextRefresh` is listed in the dep array of the final `useEffect` (the one that sets up the `visibilitychange` listener and `mountedRef`) but is never called inside the effect body. Since `scheduleNextRefresh` is stable it won't cause any extra re-runs or bugs, but it's a misleading dep — it implies the effect uses it when it doesn't. Worth removing for clarity.

**2. Inline `getIntervalMs` functions reset the poll timer on every render**

Every page passes an un-memoized inline lambda as `getIntervalMs`, for example in `Dashboard.tsx`:

```ts
getIntervalMs: () => (hasRunningSync ? DASHBOARD_FAST_REFRESH_MS : DASHBOARD_IDLE_REFRESH_MS)
```

Because this is a new function reference on every render, the hook's scheduling effect re-runs on every state change and calls `clearScheduledRefresh()` + starts a fresh timer. In practice this is harmless — renders only happen after a fetch completes, which is a natural point to reset the timer anyway. But wrapping each one in `useCallback` keyed on the relevant boolean would make the intent explicit and avoid the unnecessary timer churn.

Neither of these is blocking — both are stylistic/minor.

---
*Reviewed by [Claude Code](https://claude.ai/claude-code)*

PR #68, opened 2026-04-12
feat: add trust proxy setting to resolve X-Forwarded-For warning

### LINKED ISSUE #53: bug: review express-rate-limit X-Forwarded-For warning in Docker deployments
A warning is appearing in the Docker logs related to `express-rate-limit` and proxy handling:

```text
ValidationError: The 'X-Forwarded-For' header is set but the Express 'trust proxy' setting is false (default). This could indicate a misconfiguration which would prevent express-rate-limit from accurately identifying users.
```

This suggests Hubarr is receiving `X-Forwarded-For` headers while Express is still using the default `trust proxy = false` setting.

At the moment it is unclear whether this is:
- purely a user-side deployment/configuration issue
- something Hubarr should handle automatically or document more clearly
- a mix of both, depending on how Hubarr is deployed behind a reverse proxy

Things to investigate:
- Under what deployment conditions this warning appears
- Whether Hubarr should explicitly configure `trust proxy` for common reverse proxy setups
- Whether enabling that would have any security or behavioural implications
- Whether the current rate-limiting behaviour is incorrect when this warning is present
- Whether this should be addressed in code, documentation, or both

The goal is to determine whether this is an expected environment-specific warning or a real application-side issue that Hubarr should handle more gracefully.

Migz93:
## Summary

Adds a **Trust Proxy** toggle to General Settings that resolves the `express-rate-limit` `ValidationError` logged when Hubarr runs behind a reverse proxy. When enabled, Hubarr calls `app.set("trust proxy", true)` at startup so that rate limiting correctly identifies clients by their real IP from `X-Forwarded-For` headers rather than the proxy's IP. The setting defaults to off (no behaviour change for existing installs) and an amber **Restart Required** badge is shown in the UI since the setting is read once at process start.

The approach mirrors how Jellyseerr handles the same problem — a user-opt-in toggle rather than silently trusting all proxies, which would be a security concern in direct-access deployments.

## Changes

- **`src/shared/types.ts`** — added `trustProxy: boolean` to `AppSettings` and `SettingsResponse.general`
- **`src/server/db/settings.ts`** — added `trustProxy: false` to `defaultAppSettings` (safe default, no change for existing installs)
- **`src/server/app.ts`** — reads `trustProxy` from settings at startup and calls `app.set("trust proxy", true)` with an info log line if enabled; exposes the field via `GET /api/settings` and accepts it in `PATCH /api/settings`
- **`src/client/components/FormControls.tsx`** — added `restartRequired?: boolean` prop to `ToggleField`; renders an amber "Restart Required" badge inline next to the label when set
- **`src/client/pages/Settings.tsx`** — added Trust Proxy toggle at the top of General Settings with `restartRequired` badge and a one-sentence section description; wired into save/load

## Test plan

- [x] Open Settings → General — confirm "Trust Proxy" toggle appears at the top with the amber "Restart Required" badge and the hint text
- [x] Enable Trust Proxy and save — confirm "Saved" feedback appears with no errors
- [x] Restart the container — confirm the log shows `Trust proxy enabled — using X-Forwarded-For for client IP identification`
- [x] Disable Trust Proxy, save, restart — confirm the log line is absent and no `express-rate-limit` ValidationError appears
- [x] Confirm existing settings (Startup Sync, History Retention) are unaffected

Closes #53

🤖 Generated with Claude Code

### DIFF
--- a/src/client/components/FormControls.tsx
+++ b/src/client/components/FormControls.tsx
@@ -92,12 +92,14 @@ export function ToggleField({
   label,
   hint,
   checked,
-  onChange
+  onChange,
+  restartRequired = false
 }: {
   label: string;
   hint?: string;
   checked: boolean;
   onChange: (value: boolean) => void;
+  restartRequired?: boolean;
 }) {
   return (
     <label className="flex items-start gap-3 cursor-pointer">
@@ -120,7 +122,14 @@ export function ToggleField({
         />
       </div>
       <div>
-        <div className="text-sm font-medium text-on-surface">{label}</div>
+        <div className="flex items-center gap-2">
+          <span className="text-sm font-medium text-on-surface">{label}</span>
+          {restartRequired && (
+            <span className="text-xs font-medium px-1.5 py-0.5 rounded bg-warning/15 text-warning">
+              Restart Required
+            </span>
+          )}
+        </div>
         {hint && <div className="text-xs text-on-surface-variant mt-0.5">{hint}</div>}
       </div>
     </label>
--- a/src/client/pages/Settings.tsx
+++ b/src/client/pages/Settings.tsx
@@ -142,7 +142,7 @@ function GeneralTab({
     setError(null);
     setSuccess(false);
     try {
-      await apiPatch("/api/settings", { general: form });
+      await apiPatch("/api/settings", { general: { fullSyncOnStartup: form.fullSyncOnStartup, historyRetentionDays: form.historyRetentionDays, trustProxy: form.trustProxy } });
       setSuccess(true);
       await onSave();
     } catch (caught) {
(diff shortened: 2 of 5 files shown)


Migz93:
Reviewed this one in detail and I agree with the overall approach.

The change does accomplish the goal from #53: it gives Hubarr an explicit, user-controlled way to enable Express proxy trust when the app is deployed behind a reverse proxy, which is the right place to address the `express-rate-limit` warning about `X-Forwarded-For` handling.

I also checked the implementation path end to end and it looks solid:
- the setting is added to the shared types and default app settings
- it is exposed correctly from `GET /api/settings`
- it is accepted and persisted through `PATCH /api/settings`
- the Express app reads it during startup and applies the setting before middleware
- the UI makes the restart requirement clear, which matches how the setting is consumed

I don't see any implementation issues in the way it's been done. The default remains safe for existing installs, the behavior change is explicit rather than automatic, and the restart-required UX makes sense for a startup-time setting.

I also ran `npm run check` against the PR branch and it passed cleanly.

🤖 Generated with Codex

PR #69, opened 2026-04-12
fix: move session secret from file to database

### LINKED ISSUE #49: enhancement: review .session_secret handling and align with common app patterns
Hubarr currently creates a `.session_secret` file in the data directory and uses it for session security. This would be worth reviewing to confirm whether the current approach is the best fit long-term, and whether it matches how similar self-hosted apps handle the same problem.

Based on the current implementation:
- Hubarr generates `.session_secret` if one does not already exist
- The value is used to sign session cookies so sessions remain valid across restarts
- It can also be overridden with the `SESSION_SECRET` environment variable
- This appears to be about session signing rather than encrypting stored application data

The main concern is not necessarily that the current implementation is wrong, but that it may feel unusual compared with other apps, especially if they store the secret elsewhere, rely on environment configuration, or manage this kind of secret differently.

Things to investigate:
- How other similar self-hosted apps handle session secrets or signing keys
- Whether those apps store a generated secret in a file, environment variable, database, or another location
- Whether Hubarr's current `.session_secret` approach is appropriate as-is or should be changed
- Whether the file location should be different even if the overall approach remains valid
- Whether there are any usability, portability, backup, or security advantages to following a more common pattern

The goal is to confirm whether Hubarr's current session secret handling is the right long-term approach, and if not, identify a cleaner or more standard pattern to migrate to.

Migz93:
## Summary

Moves the session signing secret out of a standalone `.session_secret` file in the config directory and into the existing `settings` table in the SQLite database. The file-on-disk approach was flagged in #49 as feeling unusual compared to how similar self-hosted apps manage this kind of internal secret. Keeping everything in one database file is cleaner, easier to reason about for backups, and removes a loose file that users might notice or be unsure about.

The `SESSION_SECRET` environment variable override has been removed as part of this change — the database is now the single source of truth and there is no longer a mechanism to externally override the secret.

## Changes

- **`src/server/config.ts`** — removed `sessionSecret` field from `RuntimeConfig` and deleted the `resolveSessionSecret` file-based implementation entirely.
- **`src/server/db/settings.ts`** — added `"session_secret"` to `SettingKey`, added `resolveSessionSecret(db, dataDir)` which handles the one-time migration from file to DB and generates a new secret when neither exists.
- **`src/server/db/index.ts`** — calls `resolveSessionSecret` during construction and exposes `getSessionSecret()` so the app layer can access it.
- **`src/server/app.ts`** — replaces `config.sessionSecret` references with `db.getSessionSecret()` called once after `HubarrDatabase` is constructed.
- **`tests/server/test-db.ts`** — removed the now-nonexistent `sessionSecret` field from the test `RuntimeConfig` fixture.

## Test plan

- [ ] Fresh install: start the app with no existing database or config files. Confirm `hubarr.db` is created, no `.session_secret` file appears in `/config`, and the app is fully functional (login works, session persists across page reloads).
- [ ] Upgrade from existing install: place a `.session_secret` file in the data directory before starting. Confirm on first boot the file is deleted and the secret value has been migrated into the database (check via `sqlite3 hubarr.db "SELECT value FROM settings WHERE key='session_secret'"`). Confirm existing sessions remain valid (no forced re-login).
- [ ] Confirm setting `SESSION_SECRET` as an environment variable has no effect (it is no longer read).

Closes #49

### DIFF
--- a/src/server/app.ts
+++ b/src/server/app.ts
@@ -49,6 +49,7 @@ function signedValue(secret: string, value: string) {
 export function createApp(config: RuntimeConfig, scheduler?: JobScheduler) {
   const logger = new Logger(config.dataDir);
   const db = new HubarrDatabase(config);
+  const sessionSecret = db.getSessionSecret();
   const imageCache = new ImageCacheService(config.dataDir, db, logger);
   const services = new HubarrServices(db, logger, imageCache);
   const app = express();
@@ -93,7 +94,7 @@ export function createApp(config: RuntimeConfig, scheduler?: JobScheduler) {
       return next();
     }
     const [sessionId, signature] = raw.split(".");
-    if (!sessionId || !signature || signedValue(config.sessionSecret, sessionId) !== signature) {
+    if (!sessionId || !signature || signedValue(sessionSecret, sessionId) !== signature) {
       req.sessionUser = null;
       return next();
     }
@@ -110,7 +111,7 @@ export function createApp(config: RuntimeConfig, scheduler?: JobScheduler) {
   }
 
   function setSessionCookie(res: Response, sessionId: string) {
-    const signed = `${sessionId}.${signedValue(config.sessionSecret, sessionId)}`;
+    const signed = `${sessionId}.${signedValue(sessionSecret, sessionId)}`;
     res.setHeader(
       "Set-Cookie",
       `${config.sessionCookieName}=${encodeURIComponent(signed)}; HttpOnly; Path=/; SameSite=Strict; Max-Age=${Math.floor(config.sessionTtlMs / 1000)}`
--- a/src/server/config.ts
+++ b/src/server/config.ts
@@ -1,38 +1,12 @@
-import crypto from "node:crypto";
 import fs from "node:fs";
 import path from "node:path";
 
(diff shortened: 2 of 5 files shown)


Migz93:
I’ve taken another pass over this with the planned release approach in mind, and I do agree with the direction of the change.

Moving the session signing secret into the database is a cleaner default than keeping a loose `.session_secret` file beside the main app state, and the wiring through `HubarrDatabase` and the app session middleware looks sensible.

My earlier concerns were mainly around the transitional migration path and backwards compatibility. Under a normal stable-upgrade review I would have pushed harder on those, but that context changes quite a bit given the plan for the next `main` release: we are already expecting a breaking reset point and a fresh baseline.

I’ve opened #72 to track the follow-up work to flatten the migration history before release and remove transitional migration code so the migrated-to shape becomes the default. With that in mind, the migration-related rough edges in this PR feel acceptable as temporary pre-release code rather than a reason to block the change.

So my view is:

- the overall direction is good
- this does accomplish the goal of moving secret handling to a cleaner default location
- the migration concerns should be resolved as part of #72 before the next `main` release

🤖 Generated with Codex

PR #70, opened 2026-04-12
feat: improve version display for develop and branch builds

### LINKED ISSUE #48: enhancement: improve version display for develop and branch builds
The current version display logic does not seem to correctly represent non-main builds.

For example, when running the `develop` tag, the About page and the version shown in the bottom-left of the UI still report the app as the current stable release (`v1.0.2` / latest / current), which is misleading.

It would be useful to improve this so Hubarr can clearly identify when it is running:
- a stable release from `main`
- a `develop` build
- a build from another branch, such as a feature or fix branch image used for testing

Goals:
- Make the displayed version reflect the actual build source more accurately
- Distinguish stable releases from `develop` builds
- Consider whether branch-based builds should also expose their source branch in the UI
- Avoid showing a non-main build as if it were the latest stable release
- Review what build metadata is available and how it should be surfaced in the app

This should make it easier to understand exactly what version of Hubarr is running, especially during testing, development, and pre-release validation.

Migz93:
## Summary

Closes #48

Hubarr previously had no way to distinguish whether a running instance was built from `main`, `develop`, or a local dev environment — the sidebar and About page always showed `v1.0.2 / Hubarr Stable` regardless. This PR bakes build channel metadata into the Docker image at CI time and surfaces it throughout the UI.

## Changes

**CI / Docker (`Dockerfile`, `develop.yml`, `release.yml`)**
- Added `ARG BUILD_CHANNEL` and `ARG COMMIT_SHA` to the Dockerfile runtime stage, exposed as `ENV` vars (defaulting to `local` when no arg is passed)
- `develop.yml` now passes `BUILD_CHANNEL=develop` and the full commit SHA as Docker build args
- `release.yml` now passes `BUILD_CHANNEL=stable` and the full commit SHA

**Server (`src/server/version.ts`, `src/server/app.ts`)**
- `version.ts` exports `BUILD_CHANNEL` (`"stable" | "develop" | "local"`) and `BUILD_COMMIT` (7-char short SHA, or `"local"`) read from the baked-in env vars
- `/api/settings/about` now includes `buildChannel` and `commitSha` in its response

**Shared types (`src/shared/types.ts`)**
- `AboutInfo` extended with `buildChannel: "stable" | "develop" | "local"` and `commitSha: string`

**Frontend (`src/client/components/Sidebar.tsx`, `src/client/pages/Settings.tsx`)**
- Sidebar `VersionFooter` now dynamically shows:
  - **"Hubarr Stable"** + Server icon, `v1.0.x` sub-label (release builds)
  - **"Hubarr Develop"** + Beaker icon, short commit SHA sub-label (develop builds)
  - **"Local Build"** + Code2 icon, `local` sub-label (no build args, e.g. dev environment)
- Settings → About tab now shows a **Build Channel** row, and a **Commit** row (visible only on non-stable builds)

## Test plan

- [ ] Build the image locally without `--build-arg` — sidebar shows **Local Build** with Code2 icon, About tab shows channel: `local`, no Commit row
- [ ] Build with `--build-arg BUILD_CHANNEL=develop --build-arg COMMIT_SHA=abc1234567` — sidebar shows **Hubarr Develop** with Beaker icon and short SHA `abc1234`, About tab shows channel: `develop` and commit row
- [ ] Build with `--build-arg BUILD_CHANNEL=stable --build-arg COMMIT_SHA=abc1234567` — sidebar shows **Hubarr Stable** with Server icon and `v1.0.x`, About tab shows channel: `stable` with no Commit row
- [ ] Confirm the `develop` GitHub Actions workflow triggers correctly on push to `develop` and the image is tagged with the new build args
- [ ] Confirm the `release` workflow passes `stable` correctly on tag push

### DIFF
--- a/.github/workflows/develop.yml
+++ b/.github/workflows/develop.yml
@@ -60,5 +60,8 @@ jobs:
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+          build-args: |
+            BUILD_CHANNEL=develop
+            COMMIT_SHA=${{ github.sha }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -58,5 +58,8 @@ jobs:
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+          build-args: |
+            BUILD_CHANNEL=stable
+            COMMIT_SHA=${{ github.sha }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ ENV NODE_ENV=production
 ENV PORT=9301
 ENV DATA_DIR=/config
 # Build metadata — overridden by CI workflows via --build-arg
-ARG BUILD_CHANNEL=local
+ARG BUILD_CHANNEL=custom
 ARG COMMIT_SHA=local
 ENV BUILD_CHANNEL=$BUILD_CHANNEL
 ENV COMMIT_SHA=$COMMIT_SHA
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,11 @@ WORKDIR /app
 ENV NODE_ENV=production
 ENV PORT=9301
 ENV DATA_DIR=/config
+# Build metadata — overridden by CI workflows via --build-arg
+ARG BUILD_CHANNEL=local
+ARG COMMIT_SHA=local
+ENV BUILD_CHANNEL=$BUILD_CHANNEL
+ENV COMMIT_SHA=$COMMIT_SHA
 COPY --from=build /app/package.json ./package.json
 COPY --from=build /app/node_modules ./node_modules
 COPY --from=build /app/dist ./dist
--- a/src/client/components/Sidebar.tsx
+++ b/src/client/components/Sidebar.tsx
@@ -7,7 +7,9 @@ import {
   History,
   Settings,
   LogOut,
-  Server
+  Server,
+  Beaker,
+  Code2
 } from "lucide-react";
 import { apiGet } from "../lib/api";
 import { getPlexImageSrc } from "../lib/plexImage";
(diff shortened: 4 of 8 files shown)


Migz93:
A couple of follow-up points from review after discussing the intended workflow:

1. `Local Build` should be renamed to `Custom Build`.

Right now the `local` state really means “not built by the release or develop CI workflow”, which is broader than a literal local/dev build. That can include manual test images and other one-off builds as well, so `Custom Build` feels like a more accurate label for what the app is actually expressing.

2. We should explicitly *not* extend this to branch-name display.

Although the original issue mentioned branch builds, our current workflow does not publish images for arbitrary feature/fix branches, so there is no reliable branch metadata for the UI to surface. In practice, changes are tested locally first, then pushed to a branch once they are ready, so adding a separate branch-image workflow would not give us much value and would add complexity we do not currently need.

Because of that, I think the right scope here is:
- `stable`
- `develop`
- `custom build`

rather than trying to represent branch names that we do not actually build or distribute.

Aside from that, the overall direction in this PR still makes sense to me.

🤖 Generated with Codex

Migz93:
Two additional behaviour issues are still worth addressing here:

1. The sidebar currently falls back to `stable` before `/api/settings/about` loads, and also if that request fails.

That means a `develop` or custom build can still briefly, or permanently in the error path, present itself as `Hubarr Stable`. If the goal is to avoid misidentifying non-stable builds, the fallback should be neutral rather than `stable`.

2. The About page release list can still be misleading for non-stable builds.

The `Current` badge is still driven purely by `info.version` matching a GitHub release tag/name. For a `develop` build that shares the same `package.json` version as the latest stable release, that stable release can still be marked as `Current` even though the running instance is not actually the stable release image.

I think those are the last two functional gaps after the scope discussion above.

🤖 Generated with Codex

Migz93:
Thanks, this follow-up looks good to me.

The outstanding review points I raised appear to be resolved now:

- `Local Build` has been renamed to `Custom Build`, which is a better fit for any non-CI build path
- the sidebar no longer falls back to showing `Hubarr Stable` before the about API responds, so non-stable builds are not misidentified during loading or error paths
- the About page now only applies the `Current` badge for stable builds, so develop/custom builds that share the same package version no longer imply they are the current stable release image

With those changes in place, I agree with the implementation and I think it now achieves the intended goal within the scope we discussed.

🤖 Generated with Codex

Issue #54, opened 2026-04-10
enhancement: review database schema and media identifier storage


Migz93:
We should review how Hubarr stores media data in the database, with a particular focus on identifier handling and whether the current structure is causing unnecessary duplication or lookup problems.

This came up from a bug where different parts of the app were using different Plex identifier formats for what appears to be the same media item.

Example explanation from the bug investigation:
> The activity cache stores IDs as `/library/metadata/<hex>` (discover key format), but the watchlist cache stores them as `plex://movie/<hex>` or `plex://show/<hex>`. The lookup in `syncUser()` uses `item.plexItemId` which is the `plex://` format, so it never matches.

This suggests Hubarr may be storing or handling references to the same media item in multiple formats without a clear unified model for how those identifiers relate to each other.

Things to investigate:
- How media items are currently stored in the database
- Whether the same underlying item can effectively exist more than once under different identifier formats
- Whether there is unnecessary duplication across watchlist, activity, cache, or sync-related tables
- Whether each media item should have a more complete canonical record containing all known identifiers, such as discover key, `plex://` ID, GUIDs, and any other relevant references
- Whether lookups should be able to match against multiple known identifiers for the same item

Goals:
- Understand the current database structure and identifier relationships more clearly
- Identify any duplication, inconsistency, or fragmentation in stored media references
- Review whether the schema or lookup logic can be improved to reduce bugs caused by mismatched identifiers
- Consider whether a more unified media-item model would make sync, matching, and state tracking more reliable

The aim is to make media storage and matching more robust, so Hubarr can consistently recognise the same item across watchlists, activity, sync processes, and other parts of the app even when different identifier formats are involved.

Migz93:
I did a deeper review of the live Hubarr database in the running `hubarr` container while checking PR #71. PR #71 is aimed at the real item-ID mismatch and is worth fixing, but I found a few broader database issues that look separate from that PR and still belong under this issue.

What PR #71 does cover:

- `watchlist_cache` is mostly keyed by `plex://...`
- `watchlist_activity_cache` is entirely keyed by `/library/metadata/...`
- a small set of RSS rows are still stored under unstable non-canonical IDs

That mismatch is real in the live DB today.

Additional issues I found that are not really solved by PR #71:

1. Self user ID aliasing still exists

[...]

PR #71, opened 2026-04-12
fix: normalise Plex identifier formats across watchlist and activity cache

Migz93:
## Summary

Fixes the identifier format mismatch identified in #54. Two tables were storing the same Plex item under different ID formats:

- `watchlist_cache` stored items as `plex://movie/<hex>` / `plex://show/<hex>` (from the GraphQL watchlist API, which returns a `guid` field)
- `watchlist_activity_cache` stored items as `/library/metadata/<hex>` (from the `activityFeed` API, which was not requesting the `guid` field)
- RSS-sourced items fell back to an unstable compound staleKey (`<userUUID>::imdb://...|tmdb://...|tvdb://...`) when the RSS feed omitted the `plex://` GUID — which it routinely does

The hex portion is identical across all three formats — they all refer to the same underlying Plex metadata record. By normalising everything to `plex://` format, direct lookups work without fallback hacks, and the schema is consistent regardless of which source ingested the item.

## Changes

**`src/server/integrations/plex.ts`**
- Added `guid` field to the `activityFeed` GraphQL query's `metadataItem` selection (confirmed available via live API test)
- Updated `PlexActivityFeedNode.metadataItem` type to include `guid: string | null`
- Passes the GUID to `buildPlexItemId()` so new activity cache entries are stored as `plex://` rather than `/library/metadata/`

**`src/server/db/migrations.ts`** — migration v7
- Adds a `discover_key TEXT` column to `watchlist_cache`, populated from `json_extract(raw_payload, '$.discoverKey')` for all existing rows
- Normalises existing RSS staleKey `plex_item_id` values to their `plex://` GUID (extracted from the guids array in `raw_payload`). If a canonical GraphQL row already exists for the same user+item the stale RSS row is deleted; otherwise the `plex_item_id` and `discover_key` are updated in place
- Updates `image_cache` poster entries to match the new `plex_item_id` so cached poster images remain reachable
- Clears `watchlist_activity_cache` and resets the fetch job cursor so the next scheduled sync repopulates it using the corrected format

**`src/server/services.ts`**
- `processSelfRssNewItems` and `processRssNewItems` now defer `plexItemId` selection until after `enrichWatchlistItem` has run. Enrichment fetches full metadata from the Plex discover API and adds the `plex://` GUID to the item's guids array — previously the ID was set before enrichment, leaving a staleKey when the RSS feed itself didn't include the GUID
- After the plex:// GUID is resolved, `discoverKey` is also derived from it (`/library/metadata/<hex>`) and set on the item

**`src/server/db/watchlist.ts`**
- `upsertWatchlistItem` and `replaceWatchlistItems` now write `discover_key` to the new column
- `getWatchlistItems` reads `discover_key` directly from the column (with a fallback to `raw_payload` for any rows written before the migration)

## Test plan

- [ ] Rebuild and restart the container — confirm migration v7 runs cleanly (check logs for migration errors)
- [ ] Inspect `watchlist_cache`: confirm all `plex_item_id` values start with `plex://`, no staleKey rows remain, and `discover_key` is populated for every row
- [ ] Inspect `watchlist_activity_cache`: confirm the table is empty immediately after migration, then trigger an activity cache sync and verify new entries use `plex://` format
- [ ] Add a new item to the RSS watchlist feed and trigger a sync — confirm it lands with a `plex://` `plex_item_id` and a populated `discover_key`
- [ ] Run a full sync and verify `added_at` dates resolve correctly (no items stuck on the sentinel `2001-01-01` date that have a corresponding activity cache entry)
- [ ] Confirm the watchlist dashboard still displays correctly with no duplicate entries

🤖 Generated with Claude Code

### DIFF
--- a/src/server/db/migrations.ts
+++ b/src/server/db/migrations.ts
@@ -211,6 +211,92 @@ const migrations: Migration[] = [
         ALTER TABLE users ADD COLUMN collection_sort_order_override TEXT;
       `);
     }
+  },
+  {
+    version: 7,
+    up(db) {
+      // Add a dedicated discover_key column to watchlist_cache so the
+      // /library/metadata/<hex> form of each item's ID is queryable without
+      // JSON-parsing raw_payload. Populated from raw_payload on migration.
+      db.exec(`
+        ALTER TABLE watchlist_cache ADD COLUMN discover_key TEXT;
+
+        UPDATE watchlist_cache
+        SET discover_key = json_extract(raw_payload, '$.discoverKey')
+        WHERE discover_key IS NULL;
+      `);
+
+      // Normalise RSS item plex_item_ids that were stored as stableKeys
+      // (e.g. "<userUUID>::imdb://...|tmdb://...") to the canonical plex://
+      // GUID format used by all GraphQL items. The plex:// GUID is already
+      // present in the raw_payload guids array after enrichment.
+      //
+      // If a row with the target plex:// ID already exists for the same user
+      // (i.e. the item was also cached via GraphQL), the stale RSS row is
+      // deleted to eliminate the duplicate. Otherwise the plex_item_id and
+      // discover_key are updated in place.
+      //
+      // image_cache poster entries keyed on the old stableKey are updated or
+      // removed to match.
+      const rssRows = db.prepare(`
+        SELECT id, user_id, plex_item_id, raw_payload
+        FROM watchlist_cache
+        WHERE plex_item_id NOT LIKE 'plex://%'
+          AND source = 'rss'
+      `).all() as Array<{ id: number; user_id: number; plex_item_id: string; raw_payload: string }>;
+
+      const findExisting = db.prepare(
+        "SELECT id FROM watchlist_cache WHERE user_id = ? AND plex_item_id = ?"
+      );
+      const updateRow = db.prepare(
+        "UPDATE watchlist_cache SET plex_item_id = ?, discover_key = ? WHERE id = ?"
+      );
+      const deleteRow = db.prepare("DELETE FROM watchlist_cache WHERE id = ?");
+      const updateImageCache = db.prepare(
+        "UPDATE image_cache SET cache_key = ?, entity_id = ? WHERE cache_key = ?"
+      );
+      const deleteImageCache = db.prepare(
+        "DELETE FROM image_cache WHERE cache_key = ?"
+      );
+
+      db.transaction(() => {
+        for (const row of rssRows) {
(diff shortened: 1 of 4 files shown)


Migz93:
I reviewed this against the live database in the running `hubarr` container and found one blocker before merge.

The overall direction looks right: normalising `watchlist_cache` and `watchlist_activity_cache` onto `plex://...` IDs does address the core mismatch behind this bug. The problem is in migration `v7` when it rewrites poster cache keys.

On the current live DB, there are cases where the same underlying item has multiple RSS-derived poster cache rows under different old keys, but they would both be rewritten to the same canonical `poster:plex://...` key. `image_cache.cache_key` is globally unique, so the second update fails and the migration aborts with:

```text
SqliteError: UNIQUE constraint failed: image_cache.cache_key
```

I reproduced this on a copy of the live DB with `Life of a King`, where these two existing keys both target the same canonical key:

- `poster:77b5c693fcb54362::imdb://tt2708254|tmdb://201550|tvdb://23229`
- `poster:7b1d0d48c946304e::imdb://tt2708254|tmdb://201550|tvdb://23229`

Both would be rewritten to:

- `poster:plex://movie/5d9f3935d5fd3f001ee19fc7`

So I don't think this is ready to merge yet. The migration needs to handle canonical poster-key collisions safely, for example by checking whether the target poster key already exists and then either deleting the old duplicate row or merging deterministically before updating the watchlist row.

I also removed the `Closes #54` line from the PR body because this PR fixes the item-ID mismatch, but `#54` is broader and still has follow-up database issues to address separately.

🤖 Generated with Codex

Migz93:
Good catch — fixed in 6243854.

Added a `checkImageCacheKey` prepared statement that checks whether the target `poster:plex://...` key already exists in `image_cache` before attempting the rename. If it does (either from a sibling RSS row processed earlier in the same loop, or from a prior GraphQL-sourced poster fetch), we delete the stale old-key row instead of updating it, avoiding the UNIQUE constraint violation.

🤖 Generated with Claude Code

Migz93:
Re-reviewed the follow-up in `6243854` and I agree with this fix.

The new `checkImageCacheKey` guard in migration `v7` closes the blocker I raised: when a canonical `poster:plex://...` row already exists, the migration now deletes the stale old-key duplicate instead of attempting a conflicting rename, so it should avoid the `image_cache.cache_key` UNIQUE violation cleanly.

One small note only: deleting the stale `image_cache` row can leave an orphaned image file behind until cache pruning runs, but that looks non-blocking because the app already has orphan cleanup for the image cache.

From a code review perspective, I’m happy with the fix.

🤖 Generated with Codex

PR #73, opened 2026-04-13
fix: prevent blank page on invalid build channel

Migz93:
## Summary

This PR fixes a live regression where Hubarr could render a blank page after setup because the runtime reported an unexpected build channel value. It normalizes invalid build channel env values on the server, adds a defensive fallback in the client shell, and avoids noisy authenticated setup requests before a fresh install has signed in.

## Changes

- Updated the server build metadata handling in `src/server/version.ts` so any unexpected `BUILD_CHANNEL` value is normalized to `custom` instead of being returned directly to the client.
- Updated the sidebar version footer in `src/client/components/Sidebar.tsx` to fall back to the custom channel label/icon if it ever receives an unknown channel value.
- Updated onboarding flow wiring in `src/client/App.tsx` and `src/client/pages/Onboarding.tsx` so fresh installs stay on the auth step without immediately calling authenticated setup/settings endpoints.
- Rebuilt and recreated the live `hubarr` container with the existing production-style settings to verify the fix against the same runtime configuration that triggered the blank page.

## Test plan

- [x] Run `npm run check`
- [x] Run `npm run build`
- [x] Build a local `hubarr` Docker image from this branch using `BUILD_CHANNEL=local` and `COMMIT_SHA=local`
- [x] Recreate the live `hubarr` container with the existing settings: bridge networking, `/opt/hubarr:/config`, port `9301`, and `unless-stopped`
- [x] Confirm the recreated app reports a normalized build channel instead of crashing on `BUILD_CHANNEL=local`
- [x] Run `npx playwright test tests/playwright/pages.spec.ts tests/playwright/settings.spec.ts --project=chromium` against `https://hubarr.coelho.je`

🤖 Generated with Codex

### DIFF
--- a/src/client/App.tsx
+++ b/src/client/App.tsx
@@ -31,13 +31,20 @@ export default function App() {
         apiGet<BootstrapStatus>("/api/bootstrap/status"),
         apiGet<{ authenticated: boolean; user: SessionUser | null }>("/api/auth/session")
       ]);
+
       setState({
         bootstrap,
         user: session.authenticated ? session.user : null,
         loading: false
       });
+
+      return {
+        bootstrap,
+        session
+      };
     } catch {
       setState((s) => ({ ...s, loading: false }));
+      return null;
     }
   }
 
@@ -46,9 +53,12 @@ export default function App() {
   }, []);
 
   async function onAuthenticated() {
-    await loadState();
-    const updated = await apiGet<BootstrapStatus>("/api/bootstrap/status");
-    if (!updated.setupComplete) {
+    const nextState = await loadState();
+    if (!nextState) {
+      return;
+    }
+
+    if (!nextState.bootstrap.setupComplete) {
       navigate("/onboarding");
     } else {
       navigate("/dashboard");
--- a/src/client/App.tsx
+++ b/src/client/App.tsx
@@ -87,7 +87,7 @@ export default function App() {
       <Routes>
         <Route
           path="/onboarding"
-          element={<Onboarding onComplete={onSetupComplete} />}
+          element={<Onboarding authenticated={false} onComplete={onSetupComplete} />}
         />
         <Route path="*" element={<Navigate to="/onboarding" replace />} />
       </Routes>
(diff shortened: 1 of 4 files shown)


Migz93:
## Code Review Findings

Overall the PR is solid and well-scoped — the root cause is correctly diagnosed and fixed in the right place. A few things worth addressing:

---

### Issues to fix

**1. Double-fetch in `onAuthenticated` (`App.tsx:48-56`)**

`loadState()` already fetches `BootstrapStatus`, but a second `apiGet` for the same endpoint is made immediately after:

```tsx
async function onAuthenticated() {
  await loadState();                                                // fetches bootstrap + session
  const updated = await apiGet<BootstrapStatus>("/api/bootstrap/status"); // fetches bootstrap AGAIN
  ...
}
```

`loadState()` should either return its result or be refactored so the result is reusable, removing the redundant second request.

---

**2. Redundant `setStep("plex")` after `loadSetupState()` (`Onboarding.tsx:53`)**

`loadSetupState()` already calls `setStep(status.currentStep)` internally. The explicit `setStep("plex")` on the next line overrides whatever the API just returned:

```tsx
await loadSetupState();  // already calls setStep(status.currentStep)
setStep("plex");         // silently overrides the API's answer
```

If the server ever returns a step other than `"plex"` after auth, this forces the wrong step. The explicit `setStep("plex")` should be removed.

---

**3. Step 3 `done` is permanently hardcoded `false` (`Onboarding.tsx:85`)**

Steps 1 and 2 correctly render a tick icon when complete, but step 3 never will:

```tsx
<StepDot number={3} active={step === "collections"} done={false} label="Collections" />
```

The `done` prop should reflect whether collections setup has been completed, consistent with the other steps.

---

### Minor observation

The `subLabel` logic in `Sidebar.tsx:185-189` is the only channel-specific logic that lives outside `CHANNEL_CONFIG`. Not urgent, but if the channel list ever grows it would need a separate touch.

🤖 Generated with Claude Code

Migz93:
Addressed in `518460e`.

I removed the redundant bootstrap refetch after authentication, stopped overriding the server-reported onboarding step after Plex auth, and made the collections step completion state consistent with the other onboarding steps.

Verified again with `npm run check` and `npm run build`.

🤖 Generated with Codex

Issue #74, opened 2026-04-13
fix: clean up orphaned poster cache rows when watchlist items are removed


Migz93:
## Summary

Hubarr can leave behind orphaned `image_cache` poster rows after a watchlist item is removed from all watchlists.

During review of issue #54 and PR #71, the live database contained poster cache entries whose `cache_key` / `entity_id` no longer matched any current row in `watchlist_cache`. That means watchlist-owned poster metadata can outlive the watchlist items that originally caused it to be cached.

This is different from `watchlist_activity_cache`, which is expected to retain historical watchlist events even after an item is no longer on a watchlist. The cleanup concern here is specifically about watchlist-owned poster cache records and their corresponding local files.

## Problem

Right now the ownership/lifecycle rules for poster cache data are too loose:

- `watchlist_cache` is the source of truth for what is currently on any user's watchlist
- poster entries in `image_cache` are keyed by `poster:<plex_item_id>` and are used to display watchlist posters
- when a watchlist item disappears from every current watchlist, its poster cache row may remain behind
- that can leave both orphaned metadata rows and orphaned local image files until some broader cache cleanup happens

This causes avoidable cache bloat and makes the database less trustworthy, because `image_cache` can contain entries that no longer belong to any active watchlist item.

## Expected behaviour

When an item is no longer present in `watchlist_cache` for any user, Hubarr should clean up watchlist-owned poster cache data for that item.

That should include:

- deleting the orphaned `image_cache` poster row
- deleting the corresponding local cached image file if it is no longer referenced
- preserving unrelated caches and historical data such as `watchlist_activity_cache`

## Suggested fix direction

Implement explicit cleanup logic tied to watchlist replacement/removal flows.

Possible approach:

- identify poster cache keys that are no longer referenced by any `watchlist_cache.plex_item_id`
- remove their `image_cache` rows as part of watchlist reconciliation, or as a dedicated pruning step immediately after reconciliation
- remove any now-unreferenced local files for the deleted poster rows
- add logging around how many orphaned poster rows/files were removed so cleanup is observable in production

The important part is to define `image_cache` poster rows as watchlist-owned derived data, with a clear lifecycle tied to the current watchlist state.

## Notes

- This issue was split out from #54 so that #54 can stay focused on identifier modeling and alias resolution.
- Activity-cache history should remain untouched; only stale poster cache data should be cleaned up.

🤖 Generated with Codex

PR #75, opened 2026-04-13
fix: model user and media identifier aliases explicitly

Migz93:
## Summary

This PR follows up on the broader identifier-model work from #54 by making user and media identifier aliasing explicit in the database instead of relying on raw payload JSON plus service-layer fallback lookups.

The main effect is that Hubarr can now resolve the self user's numeric Plex ID and GraphQL UUID as aliases of the same local user, and can resolve the same media item across canonical `plex://...` IDs, discover keys, and external GUIDs through a dedicated identifier catalog. That makes activity-cache date resolution and future matching logic more robust and easier to reason about.

## Changes

**Database schema and backfill**
- Adds migration `v8` in `src/server/db/migrations.ts`
- Creates `user_identifier_aliases` to store multiple Plex identifiers for one local user record
- Creates `media_items` for canonical media records and `media_item_identifiers` for alias identifiers such as `plex://...`, `/library/metadata/...`, `imdb://`, `tmdb://`, and `tvdb://`
- Backfills existing `users` and `watchlist_cache` data into those new tables so current installs gain the identifier model during migration

**Identifier repository layer**
- Adds `src/server/db/identifiers.ts` with helpers to upsert user aliases, upsert media identifiers, resolve discover keys, and resolve activity-cache dates through identifier aliases
- Wires those helpers into `src/server/db/index.ts` so the service layer can use them directly

**User and watchlist persistence**
- Updates `src/server/db/users.ts` so user upserts also record identifier aliases for current Plex user IDs
- Updates `src/server/db/watchlist.ts` so watchlist writes also keep the media identifier catalog current for canonical item IDs, discover keys, and GUIDs
- Keeps `getWatchlistDiscoverKey()` compatible by resolving through the explicit identifier catalog first, then falling back to legacy payload data if needed

**Sync and activity-cache lookup flow**
- Updates `src/server/services.ts` so the self user's GraphQL UUID is stored as an alias on the same local user record during sync
- Replaces the manual “try numeric ID, UUID, item ID, discover key” activity-cache fallback chain with a single alias-aware database lookup based on local `user.id` and canonical item identity
- Registers merged watchlist items in the media identifier catalog before attempting `addedAt` resolution so newly fetched items participate in the alias model immediately

**Docs**
- Updates `docs/watchlist.md` to describe the explicit identifier tables and the new alias-based activity-cache resolution path
- Splits orphaned poster-cache cleanup into follow-up issue #74 so this PR stays focused on identifier modeling

## Test plan

- [x] Run `npm run check`
- [x] Build a local Docker image from this workspace and recreate the live `hubarr` container with the existing `/opt/hubarr:/config` bind mount, bridge networking, port `9301`, and current runtime settings
- [x] Confirm the rebuilt local container starts cleanly and listens on port `9301`
- [x] Smoke test the rebuilt local container and verify the app appears to function normally with the migrated local data
- [x] Review migration `v8` on a production-like database snapshot if deeper migration-specific verification is wanted before merge

🤖 Generated with Codex

### DIFF
--- a/docs/watchlist.md
+++ b/docs/watchlist.md
@@ -228,6 +228,19 @@ real `addedAt`. Items that remain unresolved after this lookup are recorded in
 the sync run as `watchlist.date_unresolved` and appear in the History page
 (see [Unresolved Dates in History](#unresolved-dates-in-history)).
 
+As of migration `v8`, Hubarr resolves this through explicit identifier tables
+rather than ad hoc service-layer fallbacks:
+
+- `user_identifier_aliases` stores known Plex user IDs for the same local user
+  record, including the self user's numeric Plex ID and GraphQL UUID
+- `media_items` stores one canonical `plex://...` item identity per media item
+- `media_item_identifiers` stores additional aliases for the same media item,
+  such as `/library/metadata/...` discover keys and external GUIDs
+
+That means the activity-cache lookup can match the same conceptual item and the
+same conceptual user across multiple Plex identifier formats without manually
+trying permutations in `services.ts`.
+
 ### Limitation
 
 The activity feed does not retain infinite history. Items watchlisted before the
@@ -294,6 +307,7 @@ naturally trend towards zero for most users.
 | `src/server/integrations/plex.ts` | `fetchGraphqlWatchlist()`, `fetchWatchlistActivityFeed()`, `fetchRssFeedItems()`, `fetchRssUrl()` |
 | `src/server/services.ts` | `syncUser()`, `runFullSync()`, `runUserSync()`, `syncActivityCache()`, `buildAllRssDateMaps()`, `pollRss()`, `processSelfRssNewItems()`, `processRssNewItems()`, `runPublishPass()` |
 | `src/server/db/watchlist.ts` | Watchlist and activity cache DB helpers |
+| `src/server/db/identifiers.ts` | Explicit media/user identifier catalog and alias-based lookup helpers |
 | `src/server/db/migrations.ts` | Schema — `watchlist_cache` (v1), `watchlist_activity_cache` (v5) |
 | `src/server/rss-cache.ts` | In-memory RSS diff cache — author-keyed deduplication, `prime()`, `diff()` |
 | `src/server/index.ts` | Job registrations for all three sync mechanisms |
--- a/src/server/db/identifiers.ts
+++ b/src/server/db/identifiers.ts
@@ -47,17 +47,14 @@ export function upsertMediaItemIdentifiers(
   const canonicalPlexItemId = normalizeIdentifierValue(item.plexItemId);
   if (!canonicalPlexItemId) return;
 
-  db.prepare(`
+  const mediaItem = db.prepare(`
     INSERT INTO media_items (canonical_plex_item_id, media_type)
     VALUES (?, ?)
     ON CONFLICT(canonical_plex_item_id) DO UPDATE SET
       media_type = excluded.media_type
-  `).run(canonicalPlexItemId, item.type);
-
-  const existing = db
-    .prepare("SELECT id FROM media_items WHERE canonical_plex_item_id = ?")
-    .get(canonicalPlexItemId) as { id: number } | undefined;
-  if (!existing) return;
+    RETURNING id
+  `).get(canonicalPlexItemId, item.type) as { id: number } | undefined;
+  if (!mediaItem) return;
 
   const identifiers = new Set<string>([canonicalPlexItemId]);
   if (item.discoverKey) identifiers.add(normalizeIdentifierValue(item.discoverKey));
@@ -75,7 +72,7 @@ export function upsertMediaItemIdentifiers(
   `);
 
   for (const identifier of identifiers) {
-    stmt.run(existing.id, inferMediaIdentifierType(identifier), identifier);
+    stmt.run(mediaItem.id, inferMediaIdentifierType(identifier), identifier);
   }
 }
 
--- a/src/server/db/identifiers.ts
+++ b/src/server/db/identifiers.ts
@@ -1,0 +1,136 @@
+import type Database from "better-sqlite3";
+import type { WatchlistItem } from "../../shared/types.js";
+
+type MediaIdentifierType = "plex_guid" | "discover_key" | "imdb" | "tmdb" | "tvdb" | "guid";
+type UserIdentifierType = "plex_user" | "plex_numeric" | "plex_uuid";
+
+function normalizeIdentifierValue(value: string): string {
+  return value.trim().toLowerCase();
+}
+
+function inferMediaIdentifierType(value: string): MediaIdentifierType {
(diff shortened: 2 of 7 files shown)


Migz93:
## Code Review

The overall direction is solid — moving from ad-hoc fallback chains to an explicit, queryable identifier model is the right call. A few issues worth addressing before merge.

---

### Issue 1 — Double upsert of media identifiers per sync (moderate)

In `services.ts` lines 499–501, `upsertMediaItemIdentifiers` is called for every item in `merged` before the activity cache lookup. Then at line 548, `replaceWatchlistItems` is called, which calls `upsertMediaItemIdentifiers` again for each item (via `watchlist.ts:59`).

The upserts are idempotent so nothing breaks, but the two callsites exist for different reasons that aren't documented:
- The loop at 499 is there so the identifier catalog is populated *before* the activity cache lookup at line 509
- The one inside `replaceWatchlistItems` keeps the catalog in sync with the watchlist

A future maintainer will see the pre-populate loop and the one inside the watchlist write and reasonably conclude one is redundant and delete it. A comment at line 499 explaining *why* pre-registration is needed (so the lookup on line 509 sees newly-merged items) would prevent that.

---

### Issue 2 — Extra SELECT after upsert to get the row ID (minor)

In `identifiers.ts:56–59`, after `INSERT ... ON CONFLICT DO UPDATE`, the code fires a separate `SELECT id FROM media_items WHERE canonical_plex_item_id = ?` to get the row ID. SQLite 3.35+ and `better-sqlite3` support `RETURNING`, so this could be one statement instead of two:

```ts
const existing = db.prepare(`
  INSERT INTO media_items (canonical_plex_item_id, media_type)
  VALUES (?, ?)
  ON CONFLICT(canonical_plex_item_id) DO UPDATE SET
    media_type = excluded.media_type
  RETURNING id
`).get(canonicalPlexItemId, item.type) as { id: number } | undefined;
```

The same pattern appears in `users.ts:39–44` and `users.ts:74–79` — upsert followed immediately by `SELECT id FROM users WHERE plex_user_id = ?`. All three could use `RETURNING id`.

---

### Issue 3 — `upsertSelfUser` not transaction-wrapped (minor correctness risk)

`users.ts:62–79` — the main upsert and the subsequent alias insert are two separate statements with nothing wrapping them. If the alias insert throws, the user row is persisted but the alias isn't. Compare with `upsertUsers` which correctly wraps both in `db.transaction(() => { ... })()`. `upsertSelfUser` should do the same.

---

### Issue 4 — Missing index on `identifier_value` alone (performance)

The queries in `getActivityCacheDateForUserItem` join on:

```sql
uia.identifier_value = lower(wac.plex_user_id)
-- and
mii.identifier_value = lower(wac.plex_item_id)
```

Both tables have `UNIQUE(identifier_type, identifier_value)` constraints, but SQLite's composite index can only be used when the leftmost column (`identifier_type`) is in the predicate. A lookup on `identifier_value` alone hits a full table scan of both alias tables for every row in `watchlist_activity_cache`. These indexes should have been included in the v8 migration:

```sql
CREATE INDEX idx_user_identifier_aliases_value ON user_identifier_aliases(identifier_value);
CREATE INDEX idx_media_item_identifiers_value ON media_item_identifiers(identifier_value);
```

Easy to add now before this ships — harder to fix after.

---

### Issue 5 — Duplicated helper functions between `identifiers.ts` and `migrations.ts` (maintenance concern)

`normalizeIdentifierValue`, `inferUserIdentifierType`, and `inferMediaIdentifierType` are copy-pasted verbatim in both files. The frozen-migration argument is valid, but the risk is specifically around issue 6 below: if the UUID regex in `identifiers.ts` is ever corrected, `migrations.ts` won't be updated, so backfilled rows will carry different `identifier_type` labels than going-forward rows. That would silently break alias lookups for users who installed before the fix. Worth noting as a constraint on future changes to these helpers.

---

### Issue 6 — Loose UUID regex (low)

In `identifiers.ts:22` (and the copy in `migrations.ts`):

```ts
if (/^[a-f0-9-]{32,36}$/i.test(value)) return "plex_uuid";
```

This matches any 32–36 character string of hex digits and dashes at any position. A 32-char lowercase hex string (e.g. an MD5 hash) would match as `plex_uuid`. A proper UUID is 36 chars with dashes at fixed positions:

```ts
/^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$/i
```

Practical impact is limited since numeric IDs are caught by the first branch, but the regex doesn't match the intent.

---

### Issue 7 — `getActivityCacheDateForUserItem` in `watchlist.ts` is a pointless passthrough (tidiness)

`watchlist.ts:335–341` is just:

```ts
export function getActivityCacheDateForUserItem(db, userId, plexItemId) {
  return getActivityCacheDateForUserItemFromIdentifiers(db, userId, plexItemId);
}
```

No transformation, no guard, no logic. The call chain is `index.ts → watchlistRepo → identifiersRepo` when it could be `index.ts → identifiersRepo` directly — the same way `upsertMediaItemIdentifiers` is already wired in `index.ts`. The `watchlist.ts` wrapper can be deleted.

---

### Priority before merge

Issues **4** (missing indexes — needs a follow-up migration if not added to v8 now) and **1** (double-upsert silent trap) are the ones I'd prioritise. The rest are clean-up that could go in a follow-up.

🤖 Generated with Claude Code

Migz93:
Thanks for the review. I went through the findings and pushed a follow-up commit that addresses the issues we agreed were worth taking in this PR.

Fixed in `0e937e5`:

- Issue 1: kept the pre-lookup media identifier registration, but added a code comment explaining that it is intentional and needed so newly merged items participate in alias-based activity-cache resolution before `replaceWatchlistItems()` persists the watchlist.
- Issue 2: replaced the upsert-then-select pattern with `RETURNING id` where we immediately need the inserted/updated row ID.
- Issue 3: wrapped `upsertSelfUser()` in a transaction so the self-user row and its identifier alias are written atomically.
- Issue 4: added a new migration (`v9`) with single-column indexes on `user_identifier_aliases.identifier_value` and `media_item_identifiers.identifier_value` so the new alias lookups are indexed correctly.
- Issue 7: removed the redundant `watchlist.ts` passthrough for `getActivityCacheDateForUserItem()` and wired `db/index.ts` straight to the identifiers repository.

Not fixing in this PR:

- Issue 5: the helper duplication between runtime code and migrations is a real maintenance consideration, but I’m intentionally leaving the migration copy self-contained for now. Migrations are effectively historical artifacts once shipped, so I didn’t want to add extra indirection there during this PR.
- Issue 6: the UUID regex is looser than ideal, but in this implementation alias resolution is driven by `identifier_value`, not by `identifier_type`, so it is not currently breaking lookups. Tightening it now would also raise the question of whether already-backfilled rows need relabeling, which felt like unnecessary scope for this PR.

I also rebuilt and recreated the local `hubarr` container again with the updated follow-up changes and the app came back up cleanly on the same settings for another smoke pass.

🤖 Generated with Codex

PR #77, opened 2026-04-13
fix: add maintenance tasks job for orphaned poster cache cleanup

### LINKED ISSUE #74: fix: clean up orphaned poster cache rows when watchlist items are removed

Migz93:
Closes #74

## Summary

This PR adds a dedicated daily `Maintenance Tasks` job so Hubarr can clean up orphaned watchlist-owned poster cache data automatically without tying that cleanup directly to the GraphQL reconciliation path. The new job removes stale poster metadata rows, prunes any now-unreferenced cached image files, and surfaces the maintenance work in `Settings -> Jobs` so it is visible and can be run manually.

## Changes

- Added a maintenance workflow in `src/server/services.ts` and `src/server/image-cache.ts` that deletes orphaned `poster:*` rows from `image_cache`, prunes unreferenced files from `/config/image-cache`, and logs the cleanup counts.
- Added the supporting repository and database plumbing in `src/server/db/image-cache.ts` and `src/server/db/index.ts` to identify poster cache entries whose `entity_id` no longer exists in `watchlist_cache`.
- Registered a new daily `maintenance-tasks` scheduler job in `src/server/index.ts` and exposed it through the existing Jobs API in `src/server/app.ts` so it appears in `Settings > Jobs` and supports `Run Now`.
- Added technical documentation in `docs/maintenance.md` and updated `docs/README.md`, `docs/sync.md`, and `docs/image-caching.md` so the maintenance job and its ownership model are documented for future cleanup tasks.
- Added automated coverage in `tests/server/maintenance.test.ts`, updated the Playwright settings coverage in `tests/playwright/settings.spec.ts`, and documented the new read-only check in `TESTING.md`.

## Test plan

- [x] Run `node --test --import tsx tests/server/maintenance.test.ts tests/server/watchlist-grouping.test.ts`
- [x] Run `npm run check`
- [x] Run `npm run build`
- [x] Run `npx playwright test tests/playwright/settings.spec.ts --grep "Maintenance Tasks job"`
- [x] Build a fresh `hubarr` image from this workspace and recreate the live `hubarr` container with the existing `/opt/hubarr:/config` bind mount, bridge networking, and port `9301` so the deployed instance reflects this branch for manual verification.

🤖 Generated with Codex

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -95,6 +95,7 @@ Read-only. Safe to run against a live instance.
 | Clicking a tab updates the URL | Clicks Plex, Jobs, and About tabs in turn; asserts the URL gains the expected `?tab=` parameter |
 | General tab shows Startup Sync toggle and History Retention field | Navigates to `/settings?tab=general`, waits for load, asserts both setting labels are visible |
 | Jobs tab shows the jobs table | Navigates to `/settings?tab=jobs`, asserts the "Job Name" column header is visible |
+| Jobs tab lists the Maintenance Tasks job | Navigates to `/settings?tab=jobs`, asserts the `Maintenance Tasks` row is present with its daily schedule and `Run Now` button |
 | About tab shows version and support info | Navigates to `/settings?tab=about`, asserts "About Hubarr" and "Version" headings are visible |
 | Collections tab shows watchlisted date sort options | Navigates to `/settings?tab=collections`, asserts the ordering `<select>` contains the "Watchlisted Date (New to Old)" and "Watchlisted Date (Old to New)" options |
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -16,6 +16,8 @@ long gone.
   GraphQL/RSS/activity-cache behavior, date resolution, and sync flows
 - [image-caching.md](image-caching.md) — poster/avatar caching design, stale
   refresh behavior, storage model, and cache lifecycle details
+- [maintenance.md](maintenance.md) — scheduled housekeeping tasks, current
+  cleanup responsibilities, and how to add new maintenance work safely
 
 ## When To Read Which Doc
 
@@ -27,6 +29,8 @@ long gone.
   `addedAt` handling, or collection publish triggers.
 - Read [image-caching.md](image-caching.md) when changing poster/avatar fetch,
   storage, refresh, or serving behavior.
+- Read [maintenance.md](maintenance.md) when adding background cleanup,
+  consistency checks, pruning, or other housekeeping tasks.
 
 ## Maintenance Rule
 
--- a/docs/image-caching.md
+++ b/docs/image-caching.md
@@ -146,6 +146,16 @@ Scans the cache directory and deletes any `.jpg` file whose web path is not
 present in `image_cache.local_web_path`. Safe to call at any time — only
 removes files with no matching metadata row.
 
+### `runMaintenanceTasks()`
+
+Performs image-cache housekeeping used by the daily `Maintenance Tasks` job:
+
+1. delete orphaned watchlist poster rows from `image_cache`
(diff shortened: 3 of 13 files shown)


Migz93:
## Code issue worth addressing

**`src/server/db/image-cache.ts:159-183` — Two-phase SELECT then DELETE**

The current approach selects orphaned keys outside a transaction, then deletes them one-by-one inside a transaction. This has a small but real TOCTOU window: if a sync re-adds a watchlist item between the SELECT and the DELETE, you could delete a poster entry that is now valid again. It's also more code than necessary.

A single atomic SQL statement is simpler and eliminates the race:

```typescript
export function deleteOrphanedPosterCacheEntries(db: Database.Database): number {
  const result = db.prepare(`
    DELETE FROM image_cache
    WHERE kind = 'poster'
      AND NOT EXISTS (
        SELECT 1
        FROM watchlist_cache w
        WHERE w.plex_item_id = entity_id
      )
  `).run();
  return result.changes;
}
```

`better-sqlite3`'s `.run()` returns a `RunResult` with a `.changes` property, so the count comes back directly. The SELECT and DELETE happen atomically — no intermediate array, no loop.

Everything else looks good. The layering, job registration, test coverage, and docs are all solid.

🤖 Generated with Claude Code

Migz93:
Confirmed — the new commit is exactly the right fix. The two-phase SELECT/DELETE is gone, replaced with a single atomic `DELETE ... WHERE NOT EXISTS` and `result.changes` for the count. No race window, less code. Good to go.

🤖 Generated with Claude Code

PR #79, opened 2026-04-13
chore: install Playwright Chromium deps in devcontainer example

Migz93:
## Summary

Adds a post-create setup step to the example devcontainer so Chromium and its required Linux libraries are installed automatically for Playwright-based UI testing.

## Changes

- updates `.devcontainer/devcontainer.json.example` to run `sudo npx playwright install --with-deps chromium` after the container is created
- keeps the change scoped to the tracked example config so repository users get the fix without bundling unrelated feature work

## Test plan

- [x] Validate `.devcontainer/devcontainer.json.example` is still valid JSON with `jq`
- [x] Rebuild or recreate the devcontainer so the new `postCreateCommand` runs
- [x] Re-run `npm run test:e2e` after the rebuilt container has Playwright browser deps installed

🤖 Generated with Codex

### DIFF
--- a/.devcontainer/devcontainer.json.example
+++ b/.devcontainer/devcontainer.json.example
@@ -23,6 +23,7 @@
     "HOME": "/home/node",
     "NPM_CONFIG_CACHE": "/home/node/.npm"
   },
+  "postCreateCommand": "sudo npx playwright install --with-deps chromium",
   "customizations": {
     "vscode": {
       "extensions": [
--- a/.devcontainer/devcontainer.json.example
+++ b/.devcontainer/devcontainer.json.example
@@ -23,7 +23,7 @@
     "HOME": "/home/node",
     "NPM_CONFIG_CACHE": "/home/node/.npm"
   },
-  "postCreateCommand": "sudo npx playwright install --with-deps chromium",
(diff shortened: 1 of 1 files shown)


Migz93:
The `postCreateCommand` hook is the right place for this, but there's a version mismatch risk with the current command.

`sudo npx playwright install --with-deps chromium` will pull the Playwright CLI on the fly from the npm registry — which may be a different version than the one pinned in `package.json`. At `postCreateCommand` time, `node_modules` isn't guaranteed to exist yet, so `npx` falls back to fetching from the registry.

Recommend chaining an `npm install` first so the locally-pinned version is used:

```json
"postCreateCommand": "npm install && npx playwright install --with-deps chromium"
```

This ensures you're always installing browsers for the exact Playwright version the project depends on, not whatever the registry currently serves.

🤖 Generated with Claude Code

Migz93:
Updated this to use `npm ci && npx playwright install --with-deps chromium` in the tracked example config.

That keeps the Playwright browser install aligned with the lockfile-pinned dependency version instead of letting `npx` resolve a potentially newer CLI from the registry during devcontainer setup.

🤖 Generated with Codex

Issue #80, opened 2026-04-13
bug: dashboard live-refresh Playwright test expects outdated sync label


Migz93:
## Summary

The Playwright suite has one remaining failure in `tests/playwright/live-refresh.spec.ts` after the `track-all-users` work. The failing test is not exposing a product bug in the dashboard refresh flow. Instead, it still expects the older `Publish` wording in the Recent Syncs panel while the UI now renders collection-style labels/summaries.

## Details

Failing test:
- `Dashboard recent syncs updates after a background collection sync starts and finishes`

Current failure:
- expected panel text to contain `Publish`
- actual panel text contains `Collection` / collection summary text

The dashboard appears to update correctly during and after the collection sync. The assertion just needs to be updated to match the current UI wording.

## Suggested fix

- update the test expectation in `tests/playwright/live-refresh.spec.ts`
- assert against the current dashboard label/summary shape rather than the legacy `Publish` text
- rerun `npm run test:e2e` to confirm the suite returns to green

🤖 Generated with Codex

PR #81, opened 2026-04-13
feat: track all users watchlist data

### LINKED ISSUE #56: enhancement: optionally preload watchlist data for all users
It would be useful to investigate whether Hubarr should support preloading watchlist data for all users, even if those users are not currently enabled.

The idea is that Hubarr could still collect and store watchlist data in the background, while only treating enabled users as active for dashboard visibility, publishing, and other primary workflows.

This could be exposed as an option during onboarding, such as `Preload Everyone`, and also surfaced later in settings so it can be turned on or off after setup.

Potential benefits:
- A user's watchlist data is already available as soon as they are enabled
- Admins can see which users actually use their watchlists
- It becomes easier to decide which users are worth enabling
- User-specific watchlist visibility could be shown on the users page without waiting for a later sync

Important considerations:
- Disabled-user data should not appear in the main dashboard or watchlists page by default
- Disabled-user data should not trigger collection publishing
- RSS and GraphQL sync behaviour may need adjusting so non-enabled users are still tracked without being treated as active publishing users
- We should define clearly which parts of the app are allowed to surface disabled-user data
- We should consider whether this should be opt-in only rather than enabled by default

Goals:
- Investigate support for collecting watchlist data for all users
- Add an onboarding option such as `Preload Everyone`
- Expose the same behaviour as a settings option after onboarding
- Define how disabled-user data should be stored, synced, and surfaced safely
- Keep active-user workflows separate from passive background tracking

The end result should be a more flexible setup where admins can choose to preload broader watchlist data for visibility and future use, without automatically treating every user as enabled.

Migz93:
Closes #56

## Summary

Adds a new `Track All Users` mode so Hubarr can keep background watchlist data fresh for disabled users without treating them as active publishing users. Disabled users stay out of dashboard stats, default watchlist browsing, manual per-user sync, collection publishing, and isolation, but admins can still inspect their cached watchlist data through the Users page and direct watchlist links.

This also tightens up the surrounding UX by surfacing a clear no-data state for users whose watchlists are empty or privately hidden, and by collapsing large history detail sections by default so full-sync runs are easier to review.

## Changes

- updates shared settings/types and server settings handling to add `trackAllUsers`, expose it in onboarding and Settings > General, and purge disabled-user watchlist cache when the setting is turned off
- splits sync behavior into tracked vs publishing users so full sync, RSS ingestion, and availability matching can include disabled users in the background while publish/dashboard/default watchlist flows remain enabled-only
- extends the Users page with watchlist status actions, hidden direct access to disabled-user watchlists, a no-data explanation modal, and a tidier card layout
- updates watchlist querying so disabled tracked users only appear when explicitly targeted, while normal watchlist browsing and filters remain enabled-only
- improves History detail rendering by collapsing large unmatched/error/date-unresolved sections by default
- updates technical docs for the new tracked-vs-publishing behavior and adds read-only Playwright coverage for the new settings/users UI expectations

## Test plan

- [x] `npm run check`
- [x] `npm run build`
- [x] `npm run test:e2e` 
- [x] Rebuild the live `hubarr` Docker image from this branch and recreate the live container with the same `/opt/hubarr:/config`, bridge networking, port mapping, and restart policy
- [x] Manually verify Users and History UI changes on the live instance
- [ ] Note: one unrelated Playwright assertion still needs follow-up in issue #80

🤖 Generated with Codex

### DIFF
--- a/src/client/components/PlexConfigForm.tsx
+++ b/src/client/components/PlexConfigForm.tsx
@@ -11,11 +11,13 @@ interface PlexConfigResponse {
 
 export default function PlexConfigForm({
   initialConfig,
+  initialTrackAllUsers = false,
   saveUrl,
   onSaved,
   saveLabel = "Save Plex"
 }: {
   initialConfig: PlexSettingsView | null;
+  initialTrackAllUsers?: boolean;
   saveUrl: string;
   onSaved?: (result: PlexConfigResponse) => void | Promise<void>;
   saveLabel?: string;
@@ -29,6 +31,7 @@ export default function PlexConfigForm({
   const [hostname, setHostname] = useState(initialConfig?.hostname ?? "");
   const [port, setPort] = useState(String(initialConfig?.port ?? 32400));
   const [useSsl, setUseSsl] = useState(initialConfig?.useSsl ?? false);
+  const [trackAllUsers, setTrackAllUsers] = useState(initialTrackAllUsers);
   const [saving, setSaving] = useState(false);
   const [success, setSuccess] = useState(false);
   const [error, setError] = useState<string | null>(null);
@@ -75,8 +78,13 @@ export default function PlexConfigForm({
               mode: "manual",
               hostname,
               port: Number(port),
-              useSsl
+              useSsl,
+              trackAllUsers
             };
+
+      if (payload.mode === "preset") {
+        payload.trackAllUsers = trackAllUsers;
+      }
 
       const result = await apiPost<PlexConfigResponse>(saveUrl, payload);
       setSuccess(true);
@@ -164,6 +172,13 @@ export default function PlexConfigForm({
         switchToManual();
         setUseSsl(value);
       }} />
+
+      <ToggleField
+        label="Track All Users"
+        hint="Keep background watchlist tracking running for disabled users too. Disabled users still will not publish collections or appear in normal dashboard/watchlist views."
+        checked={trackAllUsers}
+        onChange={setTrackAllUsers}
+      />
 
       <SaveBar
         saving={saving}
--- a/src/client/components/PlexConfigForm.tsx
+++ b/src/client/components/PlexConfigForm.tsx
@@ -11,13 +11,11 @@ interface PlexConfigResponse {
 
 export default function PlexConfigForm({
   initialConfig,
-  initialTrackAllUsers = false,
   saveUrl,
   onSaved,
   saveLabel = "Save Plex"
 }: {
   initialConfig: PlexSettingsView | null;
-  initialTrackAllUsers?: boolean;
   saveUrl: string;
   onSaved?: (result: PlexConfigResponse) => void | Promise<void>;
   saveLabel?: string;
@@ -31,7 +29,6 @@ export default function PlexConfigForm({
   const [hostname, setHostname] = useState(initialConfig?.hostname ?? "");
   const [port, setPort] = useState(String(initialConfig?.port ?? 32400));
   const [useSsl, setUseSsl] = useState(initialConfig?.useSsl ?? false);
-  const [trackAllUsers, setTrackAllUsers] = useState(initialTrackAllUsers);
   const [saving, setSaving] = useState(false);
   const [success, setSuccess] = useState(false);
   const [error, setError] = useState<string | null>(null);
@@ -78,13 +75,8 @@ export default function PlexConfigForm({
               mode: "manual",
               hostname,
               port: Number(port),
-              useSsl,
-              trackAllUsers
+              useSsl
             };
-
-      if (payload.mode === "preset") {
-        payload.trackAllUsers = trackAllUsers;
-      }
 
       const result = await apiPost<PlexConfigResponse>(saveUrl, payload);
       setSuccess(true);
@@ -172,13 +164,6 @@ export default function PlexConfigForm({
         switchToManual();
         setUseSsl(value);
       }} />
-
-      <ToggleField
-        label="Track All Users"
-        hint="Keep background watchlist tracking running for disabled users too. Disabled users still will not publish collections or appear in normal dashboard/watchlist views."
-        checked={trackAllUsers}
-        onChange={setTrackAllUsers}
-      />
 
       <SaveBar
         saving={saving}
--- a/src/client/pages/History.tsx
+++ b/src/client/pages/History.tsx
@@ -368,7 +368,15 @@ function RunItems({ items }: { items: SyncRunItem[] }) {
       )}
 
       {/* Match failures */}
-      {failures.length > 0 && <MatchFailuresCollapsible items={failures} />}
+      {failures.length > 0 && (
+        <ItemSectionCollapsible
+          title="unmatched items"
+          count={failures.length}
+          tone="warning"
+          items={failures}
+          renderItem={(item) => <MatchFailureRow key={item.id} item={item} />}
+        />
+      )}
 
       {/* Watchlisted at date unresolved */}
       {unresolved.length > 0 && (
@@ -485,18 +493,6 @@ function MatchFailureRow({ item }: { item: SyncRunItem }) {
         </div>
       )}
     </div>
-  );
-}
-
-function MatchFailuresCollapsible({ items }: { items: SyncRunItem[] }) {
-  return (
-    <ItemSectionCollapsible
-      title="unmatched items"
-      count={items.length}
-      tone="warning"
-      items={items}
-      renderItem={(item) => <MatchFailureRow key={item.id} item={item} />}
-    />
   );
 }
 
--- a/src/client/pages/History.tsx
+++ b/src/client/pages/History.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useState } from "react";
+import { type ReactNode, useCallback, useEffect, useState } from "react";
 import { useSearchParams } from "react-router-dom";
 import { AlertCircle, CheckCircle, ChevronDown, ChevronLeft, ChevronRight, ChevronUp, Clock, XCircle } from "lucide-react";
 import { apiGet } from "../lib/api";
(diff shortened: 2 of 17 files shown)


src/client/pages/Users.tsx
```diff
@@ -271,14 +288,20 @@ function UserCard({
...
+  onOpenWatchlist,
+  onShowNoData
 }: {
   user: UserRecord;
   selected: boolean;
   syncing: boolean;
   onToggleSelected: () => void;
   onEdit: () => void;
   onSync: () => void;
+  canOpenWatchlist: boolean;
```
Migz93: **Issue 1 — `canOpenWatchlist` prop is declared but never used**

This boolean is accepted in the interface and passed at both call sites, but is never referenced inside the `UserCard` JSX. The watchlist buttons always render based on `user.watchlistItemCount > 0` regardless of this flag — so the prop has no effect.

Either remove it, or use it to gate the watchlist navigation (e.g. prevent navigation for disabled users when `trackAllUsers` is off). As-is it's dead code.

src/server/db/users.ts
```diff
@@ -174,14 +178,23 @@ export function updateUser(
     id
   );
 
+  if (!next.enabled && !appSettings.trackAllUsers) {
```
Migz93: **Issue 2 — watchlist deletion in `updateUser` is outside any transaction**

`bulkUpdateUsers` correctly wraps both the `UPDATE users` and `DELETE FROM watchlist_cache` statements inside a single `db.transaction()`, so they're atomic. But here in `updateUser` the DELETE runs as a bare statement after the UPDATE with no wrapping transaction.

If the process crashes between these two statements, the user would be left disabled but with their watchlist cache still intact — an inconsistent state. Wrap the UPDATE + DELETE in a transaction to match the behaviour in `bulkUpdateUsers`.

src/client/pages/History.tsx
```diff
@@ -492,6 +488,18 @@ function MatchFailureRow({ item }: { item: SyncRunItem }) {
   );
 }
 
+function MatchFailuresCollapsible({ items }: { items: SyncRunItem[] }) {
```
Migz93: **Issue 3 — `MatchFailuresCollapsible` is a redundant thin wrapper**

The errors section and the date-unresolved section both call `ItemSectionCollapsible` directly at the call site. But match failures get this extra component that does nothing except re-call `ItemSectionCollapsible` with fixed args — adding an unnecessary layer and inconsistency.

Just inline `ItemSectionCollapsible` at the call site like the other two sections:
```tsx
{failures.length > 0 && (
  <ItemSectionCollapsible
    title="unmatched items"
    count={failures.length}
    tone="warning"
    items={failures}
    renderItem={(item) => <MatchFailureRow key={item.id} item={item} />}
  />
)}
```
Then `MatchFailuresCollapsible` can be removed entirely.

src/client/components/PlexConfigForm.tsx
```diff
@@ -165,6 +173,13 @@ export default function PlexConfigForm({
         setUseSsl(value);
       }} />
 
+      <ToggleField
+        label="Track All Users"
```
Migz93: **Issue 4 — `trackAllUsers` is a General setting living in the Plex form**

`trackAllUsers` is conceptually a general application setting (it lives under `general` in the settings response, not under `plex`), but it's been placed inside `PlexConfigForm` — meaning it shows up in both Settings > Plex and Settings > General, which is confusing for users.

The reason it ended up here is that onboarding uses `PlexConfigForm` and there's currently no General step in onboarding. The right fix is to **add a General onboarding step** that covers both `trackAllUsers` and `Startup Sync` (which currently only appears in Settings > General and is invisible during setup). With that step in place, `trackAllUsers` can be removed from `PlexConfigForm` entirely, and `PlexConfigPayload` no longer needs to carry a General setting through the Plex save handler in `app.ts`.

src/server/services.ts
```diff
@@ -260,6 +260,15 @@ export class HubarrServices {
     };
   }
 
+  private listTrackedUsers() {
```
Migz93: **Issue 5 — redundant DB reads from `listTrackedUsers` and `listPublishingUsers`**

Each helper makes its own full DB round-trip:
```typescript
private listTrackedUsers() {
  const settings = this.db.getAppSettings(); // DB read
  return this.db.listUsers().filter(...);     // DB read
}
private listPublishingUsers() {
  return this.db.listUsers().filter(...);     // DB read
}
```
`runFullSync` calls both, resulting in 3 user-related DB reads where 1 would suffice. Consider a single helper that returns `{ tracked, publishing }` from one `listUsers()` call, or have callers pass in an already-fetched users array.

src/server/services.ts
```diff
@@ -1279,8 +1297,8 @@ export class HubarrServices {
       return 0;
     }
 
-    if (!selfUser.enabled) {
-      this.logger.debug("Self user is not enabled, skipping self RSS items");
+    if (!this.listTrackedUsers().some((user) => user.id === selfUser.id)) {
```
Migz93: **Issue 6 — self-user RSS guard is over-engineered**

```typescript
if (!this.listTrackedUsers().some((user) => user.id === selfUser.id))
```
This fetches **all users + settings** from the DB on every RSS poll cycle just to check whether the self user is currently tracked. A simpler equivalent that avoids the extra reads:
```typescript
const settings = this.db.getAppSettings();
if (!selfUser.enabled && !settings.trackAllUsers)
```

Note: `selfUser.enabled` is not always true (fresh-install state), so the `selfUser.enabled` check is still needed — but `listTrackedUsers().some()` is overkill for checking a single known user. The simplified form handles all cases correctly:
- Disabled self + `trackAllUsers` off → skip ✓  
- Disabled self + `trackAllUsers` on → process RSS ✓  
- Enabled self → process RSS ✓

src/server/db/watchlist.ts
```diff
@@ -109,6 +110,22 @@ export function getWatchlistGrouped(
 ): WatchlistPageResponse {
   const { userId, mediaType, availability, sortBy = "added-desc", page, pageSize } = options;
   const offset = (page - 1) * pageSize;
+  const appSettings = getAppSettings(db);
```
Migz93: **Issue 7 — `getAppSettings()` called inside `getWatchlistGrouped()` on every request**

`getWatchlistGrouped` is called on every watchlist page load. This adds an extra DB read to every request. For the common case where `userId` is `undefined` (normal browsing), `allowSelectedDisabledOnly` is always `false` and the settings read is wasted.

Consider only reading settings when a `userId` is actually provided:
```typescript
const allowSelectedDisabledOnly = Boolean(
  selectedUser && !selectedUser.enabled &&
  userId && getAppSettings(db).trackAllUsers
);
```
Or move the settings lookup to the call site and pass it in as a parameter.

src/client/pages/Users.tsx
```diff
@@ -328,13 +374,57 @@ function UserCard({
...
+  onClose: () => void;
+}) {
+  const detail = !user.enabled && !trackAllUsersEnabled
+    ? "No watchlist data is available for this user right now. They are currently disabled in Hubarr and Track All Users is turned off."
+    : "No watchlist data can be found for this user. This usually means their watchlist privacy is set to private or they currently have no items in their watchlist.";
+
+  return (
+    <div
+      className="fixed inset-0 bg-background/80 backdrop-blur-sm z-50 flex items-center justify-center p-4"
+      onClick={onClose}
```
Migz93, line 405: **Issue 8 — `WatchlistInfoModal` has no keyboard dismissal (Escape key)**

Clicking the backdrop closes the modal, but there's no `onKeyDown` / `onKeyUp` handler for the Escape key. This is a minor accessibility gap — users expect modals to be dismissable with Escape. A simple addition to the outer `div`:
```tsx
onKeyDown={(e) => { if (e.key === 'Escape') onClose(); }}
```

Migz93 (commented):
Good feature overall — the tracked vs. publishing user split is clean and consistently applied. Eight issues below, ranging from a dead prop to a transaction gap. The two items we discussed further are issues 4 and 6.

Migz93:
## Review follow-up

All 8 issues from the review have been addressed in the fix commit. Summary of what was resolved:

**Fixed by Codex:**

1. **`canOpenWatchlist` dead prop** — removed from `UserCard` interface and both call sites
2. **`updateUser` transaction gap** — UPDATE and DELETE now wrapped in a single `db.transaction()`, consistent with `bulkUpdateUsers`
3. **`MatchFailuresCollapsible` redundant wrapper** — removed; call site now uses `ItemSectionCollapsible` directly inline like the other two sections
4. **`trackAllUsers` in Plex form** — removed from `PlexConfigForm` and `PlexConfigPayload` entirely; a new General step (step 3 of 4) added to onboarding covering both `trackAllUsers` and `Startup Sync`
5. **Redundant DB reads** — `listTrackedUsers`/`listPublishingUsers` replaced with a single `getUserScopes()` call returning `{ settings, users, trackedUsers, publishingUsers }`; `processRssNewItems` now receives `trackedUsers` as a parameter
6. **Self-user RSS guard** — simplified to `!selfUser.enabled && !settings.trackAllUsers`, eliminating the unnecessary `listTrackedUsers().some()` full-fetch
7. **`getAppSettings` on every watchlist load** — top-level call removed; now only executes when a `userId` is provided and the selected user is disabled (short-circuits via `&&`)
8. **Missing Escape key on `WatchlistInfoModal`** — `useEffect` added with `document.addEventListener("keydown", ...)` and proper cleanup

**Fixed in this follow-up commit (`9ae9802`):**

- **Double `getAppSettings` read in `pollRss`** — `settings` is now destructured from the existing `getUserScopes()` call at the top of `pollRss` and passed into `processSelfRssNewItems`, removing the internal `this.db.getAppSettings()` call from that method. One DB read per RSS poll cycle instead of two.
- **`TESTING.md` missing onboarding documentation** — added an Onboarding flow section documenting the four steps (Auth → Configure Plex → General → Collections) and noting that step order is enforced server-side by `getCurrentOnboardingStep`.

PR #85, opened 2026-04-13
fix: fix Plex OAuth popup not closing on mobile through reverse proxy

### LINKED ISSUE #83: bug: Plex auth popup does not close correctly on mobile when Hubarr is accessed through a proxy
When logging into Hubarr on mobile through a reverse proxy, the Plex authentication popup does not close correctly after authentication completes.

Observed behaviour:
- The Plex login/auth window stays open instead of closing automatically
- If the popup is manually closed and focus is returned to Hubarr, Hubarr reports that the popup was closed too early

Comparison:
- When Hubarr is accessed directly via `IP:PORT`, the Plex authentication flow works correctly
- When Hubarr is accessed through a proxy on desktop, there is no issue
- The issue appears to happen specifically on mobile when Hubarr is accessed through a proxy

This suggests there may be an issue with how the popup flow, redirect handling, origin checks, or callback handling behaves in a mobile browser behind a reverse proxy.

Things to investigate:
- Whether the proxy changes the callback/origin behaviour used by the Plex auth popup flow
- Whether Hubarr is constructing any URLs incorrectly when accessed through a proxied hostname
- Whether mobile browser popup behaviour differs from desktop in a way that affects the auth flow
- Whether browser security/origin rules are affecting popup close or postMessage-style communication
- Whether this is related to reverse proxy headers or trusted proxy handling

Expected behaviour:
The Plex auth popup should close automatically after successful authentication, and Hubarr should complete the login flow normally on mobile as well as desktop, whether the app is accessed directly or through a proxy.

Migz93:
## Summary

This PR fixes the Plex OAuth flow on mobile by making Hubarr handle the popup more like Seerr does, while also keeping the extra polling grace period that protects against Plex token timing races.

Instead of opening `about:blank` and immediately relying on the opener to close a cross-origin Plex tab, Hubarr now opens a same-origin loading page first, waits briefly so mobile browsers treat it as a user-opened popup/tab, redirects to Plex for authentication, and then sends Plex back to a Hubarr-owned completion page that closes itself. The result is that both desktop and mobile complete sign-in cleanly, including the mobile case where the Plex tab previously stayed open after successful login.

## Changes

- `src/client/lib/plexOAuth.ts`
  - Opens the Plex auth popup to `/login/plex/loading` instead of `about:blank`, focuses it, and keeps the short `1500ms` delay before redirecting to Plex so mobile browsers preserve the popup/tab correctly.
  - Adds a clearer error when the popup cannot be opened at all.
  - Passes `forwardUrl: ${window.location.origin}/login/plex/done` to Plex so the auth window returns to a Hubarr-owned page that can close itself reliably.
  - Keeps the extra post-close polling grace period so Hubarr does not reject logins if Plex returns the auth token a moment after the popup closes.

- `src/client/App.tsx`
  - Adds a lightweight public `/login/plex/loading` route that shows a spinner-only loading state while the popup is being handed off to Plex.
  - Adds a lightweight public `/login/plex/done` route that attempts to close the popup from inside the Hubarr-owned page after Plex redirects back.

## Test plan

- [x] Desktop / direct access: click Plex login, authenticate, confirm the popup closes and Hubarr finishes sign-in.
- [x] Desktop / proxied access: confirm the same flow works without regression.
- [x] Mobile / direct access: click Plex login, confirm the spinner page appears briefly, authenticate, and confirm the Plex tab closes and Hubarr ends up signed in.
- [x] Mobile / proxied access: confirm the same flow works when accessing Hubarr via the proxied hostname.
- [x] Confirm the previous mobile failure mode is gone: after successful Plex auth, the Plex window/tab no longer stays open and Hubarr no longer reports that the popup was closed too early.

Closes #83

🤖 Generated with Codex

### DIFF
--- a/src/client/App.tsx
+++ b/src/client/App.tsx
@@ -20,12 +20,6 @@ interface AppState {
 
 export default function App() {
   const location = useLocation();
-  const navigate = useNavigate();
-  const [state, setState] = useState<AppState>({
-    bootstrap: null,
-    user: null,
-    loading: true
-  });
 
   // Keep the Plex popup on a lightweight same-origin page so mobile browsers
   // treat the auth window as a user-opened tab before it navigates to plex.tv.
@@ -36,6 +30,17 @@ export default function App() {
   if (location.pathname === "/login/plex/done") {
     return <PlexPopupDone />;
   }
+
+  return <MainApp />;
+}
+
+function MainApp() {
+  const navigate = useNavigate();
+  const [state, setState] = useState<AppState>({
+    bootstrap: null,
+    user: null,
+    loading: true
+  });
 
   async function loadState() {
     try {
@@ -180,7 +185,7 @@ function PlexPopupDone() {
 
   return (
     <div className="min-h-screen bg-background flex items-center justify-center p-4">
-      <RefreshCw size={28} className="animate-spin text-primary" aria-label="Loading" />
+      <p className="text-on-surface-variant text-sm">Authentication complete. You can close this tab.</p>
     </div>
   );
 }
--- a/src/client/App.tsx
+++ b/src/client/App.tsx
@@ -1,5 +1,6 @@
 import { useEffect, useState } from "react";
-import { Navigate, Route, Routes, useNavigate } from "react-router-dom";
+import { Navigate, Route, Routes, useLocation, useNavigate } from "react-router-dom";
+import { RefreshCw } from "lucide-react";
 import { apiGet, apiPost } from "./lib/api";
 import Layout from "./components/Layout";
 import Login from "./pages/Login";
@@ -18,12 +19,23 @@ interface AppState {
 }
 
 export default function App() {
+  const location = useLocation();
   const navigate = useNavigate();
   const [state, setState] = useState<AppState>({
     bootstrap: null,
     user: null,
     loading: true
   });
+
+  // Keep the Plex popup on a lightweight same-origin page so mobile browsers
+  // treat the auth window as a user-opened tab before it navigates to plex.tv.
+  if (location.pathname === "/login/plex/loading") {
+    return <PlexPopupLoading />;
+  }
+
+  if (location.pathname === "/login/plex/done") {
+    return <PlexPopupDone />;
+  }
 
   async function loadState() {
     try {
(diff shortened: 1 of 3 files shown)


Migz93:
## Code Review

Overall the approach is correct and well-thought-out. One change recommended before merging, plus a few minor observations.

---

### Issues

**1. `PlexPopupDone` shows a spinner on failure — UX issue**

In `PlexPopupDone` ([App.tsx](src/client/App.tsx)), if both `window.close()` calls fail (which happens when the page wasn't opened via JS — e.g. a user manually navigating to `/login/plex/done`), the user is left staring at an infinite spinner with no context. The spinner implies something is still loading, when actually the only thing left to do is close the window.

Recommend swapping the spinner for a fallback message:

```tsx
return (
  <div className="min-h-screen bg-background flex items-center justify-center p-4">
    <p className="text-on-surface-variant text-sm">Authentication complete. You can close this tab.</p>
  </div>
);
```

**2. Unguarded `response.ok` in `pollForToken` — pre-existing, worth a follow-up**

In `plexOAuth.ts`:
```ts
const response = await fetch(`https://plex.tv/api/v2/pins/${this.pin.id}`, { headers: this.headers });
const data = (await response.json()) as { authToken?: string | null };
```
If Plex returns a non-200 (e.g. expired PIN), it will attempt to parse the JSON regardless. Not introduced by this PR but worth a follow-up issue.

---

### Minor observations (non-blocking)

- **`gracePollsLeft--` in a condition** — post-decrement inside a condition is a minor code smell. `gracePollsLeft > 0` followed by `gracePollsLeft -= 1` inside the block would read more clearly, but it works correctly as-is.
- **`useNavigate()` called but unused in popup routes** — when the early returns for `/login/plex/loading` or `/login/plex/done` fire, `navigate` is never used. Not a bug (hooks rules are satisfied), just a minor wasteful hook call.
- **Redundant null-check at `if (this.popup)` before setting `location.href`** — always truthy at that point since the check at the top of `login()` throws if the popup is missing. Harmless but redundant.

---

### What's good

- The core mechanism (same-origin loading page → Plex → same-origin done page that self-closes) is the right fix and mirrors how Overseerr handles it.
- `screenLeft ?? screenX` fallback is a solid cross-browser improvement.
- Grace polling (5 extra polls after popup closes) is a sensible guard against Plex token delivery lag.
- Comments throughout explain the *why*, not just the *what*.
- The 1500ms delay is well-justified in the comment.

**Bottom line:** Ready to merge once the `PlexPopupDone` spinner is swapped for a "you can close this tab" message. Everything else is non-blocking.

🤖 Generated with Claude Code

Migz93:
Addressed the review feedback in a follow-up patch:

- kept the spinner-only loading page before Plex auth
- changed the post-auth done page to a clear "Authentication complete. You can close this tab." fallback message
- added `response.ok` handling in the PIN poller
- cleaned up the grace-poll decrement logic and removed a redundant popup null-check
- split the popup routes away from the main app flow so we don't mount the main navigation hook for those lightweight pages

The updated code is pushed on this branch now.

🤖 Generated with Codex

Migz93:
All feedback from the previous review has been addressed. Good to merge.

| Feedback | Status |
|---|---|
| `PlexPopupDone` spinner → "Authentication complete. You can close this tab." | Fixed |
| Unguarded `response.ok` in `pollForToken` | Fixed — now throws on non-200 |
| `gracePollsLeft--` in condition → cleaner two-step decrement | Fixed |
| `useNavigate()` unused in popup routes | Fixed — refactored into a `MainApp` sub-component, cleanly separating the popup routes from the main app logic |
| Redundant `if (this.popup)` null-check | Fixed — removed, now calls directly |

The `MainApp` refactor is a nicer solution than suggested — rather than just moving a hook call, it properly separates concerns so the popup pages are fully lightweight and the main app state/navigation logic is isolated in its own component.

🤖 Generated with Claude Code

PR #86, opened 2026-04-14
feat: improve history job progress updates

### LINKED ISSUE #84: enhancement: show in-progress job progress for GraphQL and collection tasks
It would be useful to show more detailed progress information while longer-running jobs are active, especially on the `History` page.

For jobs such as GraphQL syncs and collection updates, it would be helpful to show progress like:

`Working on user X out of Y`

That would make it much easier to understand how far through the user list the job has progressed, rather than only seeing that the job is running.

Likely candidates:
- GraphQL sync jobs
- Collection jobs

This probably does not apply to RSS feed work, since that does not appear to process users in the same way, but this should be reviewed as part of the investigation.

Goals:
- Show useful in-progress status updates for long-running jobs on the `History` page
- Indicate which user is currently being processed and how far through the total list the job is
- Improve visibility into job progress without needing to rely only on start/end states
- Review whether any other job types are also suitable for this kind of progress display

The end result should be that when a supported job is running, the `History` page gives a clearer sense of what Hubarr is currently doing and how far through the task it is.

Migz93:
Closes #84

## Summary

This PR improves the History page feedback for long-running GraphQL and collection jobs by updating the running row summary as each user is processed, so the page now shows which user is currently being worked on and how far through the run Hubarr is.

As we tested the issue, we also made a couple of related UX improvements that were not called out explicitly in the original issue but ended up being the better experience in practice. Expanded History rows now keep refreshing while a run is active, and pages that already use live polling continue updating even when their tab is in the background. That means job state, recent syncs, and history activity stay current without needing to collapse/reopen panels or switch back and press refresh.

## Changes

- Added a server-side `sync_runs.summary` update path so running full-sync and collection-sync jobs can publish per-user progress text while they are still active.
- Updated the full GraphQL sync and collection publish loops to write live in-progress summaries for the current user and current/total position.
- Updated the History page so expanded run details poll for fresh data while a run is still in progress, letting the summary, status, timestamps, errors, and step details update without a page reload.
- Extended the shared client live-refresh hook with an opt-out for pausing in hidden tabs, then enabled background polling on History, expanded History details, Dashboard recent syncs, Jobs, Users, and Watchlists.
- Added a focused server regression test for running-summary updates and a read-only Playwright spec that verifies History background polling with stubbed API responses.
- Documented the new Playwright coverage in `TESTING.md`.

## Test plan

- [x] Run `npm run check`
- [x] Run `npx tsx --test tests/server/history-progress.test.ts`
- [x] Run `npx playwright test tests/playwright/history-background-refresh.spec.ts --project=chromium`
- [x] Rebuild and restart the live `hubarr` container, then verify on the real app that GraphQL and Collection runs show live per-user progress on the History page.
- [x] Verify expanded History rows keep updating while a run is active without needing a full page refresh.
- [x] Verify the History page, Jobs tab, Dashboard recent syncs, and other polling views continue updating while their tab is in the background.

🤖 Generated with Codex

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -153,3 +153,16 @@ without a browser reload. They are not read-only.
 | Dashboard recent syncs updates after a background collection sync starts and finishes | Opens `/dashboard`, triggers the collection-sync job through the API, and verifies the `Recent Syncs` panel first shows a running Publish entry and then updates to the finished summary without a reload |
 | History shows a new collection sync row move from running to its terminal status without reload | Opens `/history`, triggers the collection-sync job through the API, and verifies the newest row appears as running and then changes to its final status/summary automatically |
 | Jobs shows a scheduler-managed job running and then returning to Run Now after polling catches completion | Opens `Settings > Jobs`, clicks `Run Now` for `Collection Sync`, verifies the row enters a running state, and then returns to idle with an updated terminal status once the job finishes |
+
+---
+
+### `tests/playwright/history-background-refresh.spec.ts` — History background polling
+
+Read-only. Safe to run against a live instance. These tests stub the History API
+responses in-browser so they can verify polling behavior without triggering real
+jobs on the server.
+
+| Test | What it checks |
+|---|---|
+| History list keeps polling while the tab is hidden when a run is active | Opens `/history` with a stubbed running sync, switches to another tab, and verifies the hidden History page still polls `/api/history` |
+| Expanded history details keep polling while the tab is hidden when a run is active | Expands a stubbed running History row, switches to another tab, and verifies the hidden page keeps polling `/api/history/:runId` for updated details |
--- a/src/client/lib/useLiveRefresh.ts
+++ b/src/client/lib/useLiveRefresh.ts
@@ -3,16 +3,18 @@ import { useCallback, useEffect, useRef } from "react";
 interface LiveRefreshOptions {
   enabled?: boolean;
   getIntervalMs: () => number | null;
+  pauseWhenHidden?: boolean;
 }
 
 export function useLiveRefresh(
   load: () => Promise<void>,
-  { enabled = true, getIntervalMs }: LiveRefreshOptions
+  { enabled = true, getIntervalMs, pauseWhenHidden = true }: LiveRefreshOptions
 ) {
   const mountedRef = useRef(true);
   const loadRef = useRef(load);
   const getIntervalMsRef = useRef(getIntervalMs);
   const enabledRef = useRef(enabled);
+  const pauseWhenHiddenRef = useRef(pauseWhenHidden);
   const visibleRef = useRef(typeof document === "undefined" ? true : document.visibilityState === "visible");
   const inFlightRef = useRef(false);
   const timeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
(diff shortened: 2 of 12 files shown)


Migz93:
## Code Review

Overall this is a solid, well-structured PR. The core approach is correct and the implementation is clean. A few things worth noting:

---

**`setDetailLoading` is never cleared on a failed initial load**

In `RunRow.loadDetail`, if the non-background (initial expand) fetch throws, `setDetailLoading(false)` still fires in the `finally` block so the spinner clears correctly. However `detail` remains `null`, leaving the expanded panel empty with no error message and no way for the user to know the fetch failed. This is a pre-existing gap that the refactor didn't address — worth a follow-up rather than blocking this PR.

**Progress summary is written at the start of each user's iteration**

In `services.ts`, `updateRunProgressSummary` is called before the work for each user begins, so the last user will show as `(N/N)` while the job is still processing them. This is the standard "currently processing" pattern and is fine UX — just worth being conscious of when reading the summary text.

**Playwright tests use a hard `waitForTimeout(6_000)` sleep**

Both new tests sleep for 6 seconds and then assert the request count increased. This works but could be fragile on a slow CI runner if the polling interval lines up poorly. A more resilient pattern would be `expect.poll(() => requestCount).toBeGreaterThan(initial, { timeout: 10_000 })` rather than a fixed sleep. Low priority since the tests are read-only and won't affect the server.

**SQL guard on `updateSyncRunSummary` is good**

The `WHERE status = 'running'` clause prevents overwriting a completed run's summary if something calls the function late. The server test explicitly covers this case — nice.

**`liveRun = detail ?? run` pattern is clean**

Collapsed rows still show live list-level data, expanded rows upgrade to full detail data without any extra branching. Good approach.

---

## Separate — `pauseWhenHidden` can probably be removed entirely

After reviewing the PR we checked whether any call site actually uses the default `pauseWhenHidden: true`. None do — every page passes `false`. Since the option exists only to set itself to its non-default value everywhere, it's worth simplifying this in a follow-up:

- Remove `pauseWhenHidden` from the `LiveRefreshOptions` interface in `src/client/lib/useLiveRefresh.ts`
- Remove the `pauseWhenHidden = true` destructure parameter
- Remove `pauseWhenHiddenRef` and all references to it
- In `scheduleNextRefresh`, remove the `if (pauseWhenHiddenRef.current && !visibleRef.current)` guard — the check for `!visibleRef.current` goes away entirely
- In the `visibilitychange` handler, remove the `else if (pauseWhenHiddenRef.current)` branch that cancels the schedule on hide
- Drop `pauseWhenHidden: false` from all 6 call sites (Dashboard, History ×2, Settings/Jobs, Users, Watchlists)

The only reason to keep the option would be if a future page needed to opt back in to pausing — but given every page in the app benefits from background updates, that seems unlikely. Removing it keeps the hook simpler and removes the question of which value to pass for new call sites.

🤖 Generated with Claude Code

Migz93:
Addressed the follow-up items from review in the latest push:

- Added a visible error state plus a retry action when expanding a History row fails to load its detail payload, instead of leaving the panel blank.
- Reworked the new read-only Playwright background-refresh spec to use `expect.poll(...)` rather than fixed `waitForTimeout(...)` sleeps.
- Simplified the shared client live-refresh hook by removing `pauseWhenHidden`, since every current polling page now refreshes in background tabs.

Re-ran `npm run check` and `npx playwright test tests/playwright/history-background-refresh.spec.ts --project=chromium`, and both passed.

🤖 Generated with Codex

PR #88, opened 2026-04-14
feat: improve history job detail display

### LINKED ISSUE #87: enhancement: review and improve History page job detail display
The `History` page would benefit from a broader review of how job details are displayed.

At the moment, some information is too noisy, some useful information is missing, and some wording/presentation could be clearer and easier to scan.

Areas to review:

1. Error message display
- Error messages should be collapsed by default
- At the moment, long error text takes up a lot of space as soon as the top-level item is expanded
- It looks like the error is already available in its own collapsed section, so the duplicate or automatically expanded display should be reconsidered

2. GraphQL job step display
- GraphQL jobs currently show many repeated `Watchlist fetch` rows, which appear to reflect pagination rather than meaningful user-facing steps
- Instead of showing each pagination fetch as a separate row, these should probably be grouped into a single logical `Watchlist fetch` step
- More broadly, GraphQL jobs should show the actual high-level steps they perform, not just one narrow category of step
- For example, useful steps might include watchlist fetch, processing users, and launching collection syncs if applicable

3. Subtext formatting
- The short summary text under each event should start with a capital letter
- Examples like `finished for 40 users.` or `finished for self.` would read better as `Finished for 40 users.` and `Finished for self.`
- More generally, review history page text to ensure capitalization is consistent wherever it makes sense

4. Running job timing
- While jobs are running, the time display can sit on `just now` for too long without adding much useful information
- It would be helpful to show elapsed time while a job is still active, for example:
  - `Just Now`
  - `Running for 42 seconds`
- `Just Now` should also be capitalized

5. Collection sync step detail
- Collection sync currently shows repeated generic `collection publish` steps
- It would be more useful if each step identified which collection was actually being published
- That would make the job history much easier to understand at a glance

6. RSS sync step detail
- When RSS sync finds nothing, it currently shows `No step details recorded.`
- That is misleading if the job did run meaningful checks but simply found no new items
- Instead, it should show what it checked, such as self RSS feed and other RSS feeds where relevant
- When RSS items are found, the step details should be more descriptive, for example identifying which item was found for which user rather than only showing a generic label such as `RSS item (self)`

Goals:
- Make the `History` page easier to scan and less visually noisy
- Show more meaningful and accurate step detail for each job type
- Improve wording, capitalization, and presentation consistency
- Surface more useful context about what each job actually did and what it found
- Improve running-job visibility with elapsed time and clearer live status

The end result should be a `History` page that feels more polished, more informative, and more useful for understanding what Hubarr is doing, what happened during a job, and where issues occurred.

Migz93:
Closes #87

## Summary

This PR upgrades the History page so job runs are easier to scan and the detail view reflects the real work each run performed instead of mostly dumping low-level raw events. It improves subtitle/status wording, adds clearer live-running timing, collapses errors by default, and makes GraphQL, RSS, and collection history rows more descriptive and actionable.

## Changes

- Updated the History page UI in  to normalize raw run items into grouped display steps, capitalize displayed summaries/statuses, show live  timing, collapse error sections by default, and render per-user watchlist fetch rows instead of one combined GraphQL fetch summary.
- Enriched history payloads in  so full sync, collection publish, and RSS runs record the extra context the UI now needs, including user display names, feed-check detail, collection publish follow-up events, and clearer per-step metadata.
- Adjusted shared presentation helpers in , updated sync/testing docs in  and , and refreshed Playwright coverage in the History-related specs so the tests assert the new wording and grouped-step behavior.

## Test plan

- [x] Run 
> hubarr@1.0.2 check
> tsc --noEmit -p tsconfig.json && tsc --noEmit -p tsconfig.server.json
- [x] Run 
> hubarr@1.0.2 build
> npm run build:client && npm run build:server

> hubarr@1.0.2 build:client
> vite build

vite v8.0.8 building client environment for production...
[2K
transforming...✓ 1748 modules transformed.
rendering chunks...
computing gzip size...
dist/client/index.html                   0.66 kB │ gzip:   0.38 kB
dist/client/assets/index-BPikDTI0.css   39.60 kB │ gzip:   7.39 kB
dist/client/assets/index-C1fGOHHD.js   354.61 kB │ gzip: 101.57 kB

✓ built in 269ms

> hubarr@1.0.2 build:server
> tsc -p tsconfig.server.json
- [x] Run the History-focused Playwright specs after updating assertions for the new UI
- [x] Run a full Playwright pass against the rebuilt live Hubarr instance
- [x] Perform a manual UI sanity pass on the History page using the rebuilt live container

🤖 Generated with Codex

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -164,5 +164,8 @@ jobs on the server.
 
 | Test | What it checks |
 |---|---|
+| Running history rows show `Just now` and live elapsed text | Opens `/history` with a stubbed running sync and verifies the row shows capitalized relative time plus a live `Running for ...` duration |
 | History list keeps polling while the tab is hidden when a run is active | Opens `/history` with a stubbed running sync, switches to another tab, and verifies the hidden History page still polls `/api/history` |
 | Expanded history details keep polling while the tab is hidden when a run is active | Expands a stubbed running History row, switches to another tab, and verifies the hidden page keeps polling `/api/history/:runId` for updated details |
+| Expanded errors stay collapsed by default and grouped steps render readable labels | Opens a stubbed failed History run, verifies the errors section stays collapsed initially, and checks that repeated low-level steps render as grouped human-readable labels |
+| RSS runs show feed checks and descriptive item labels | Opens a stubbed RSS History run and verifies feed-check steps and found-item labels render instead of a generic empty-details state |
--- a/docs/sync.md
+++ b/docs/sync.md
@@ -220,7 +220,10 @@ History stores higher-level sync runs such as:
 - `publish`
 
 It also stores per-run detail rows for actions such as watchlist fetches, match
-failures, unresolved `addedAt` dates, and collection updates.
+failures, unresolved `addedAt` dates, collection updates, and RSS feed checks.
+The History page groups those raw rows into more readable operational steps so
+the UI reflects what a job actually did rather than dumping every low-level
+event verbatim.
 
 ### Jobs
 
--- a/src/client/lib/utils.ts
+++ b/src/client/lib/utils.ts
@@ -40,7 +40,7 @@ export function formatRelativeTime(isoString: string): string {
   const diffHours = Math.floor(diffMins / 60);
   const diffDays = Math.floor(diffHours / 24);
 
-  if (diffSecs < 60) return "just now";
+  if (diffSecs < 60) return "Just now";
 
   if (diffMs > 0) {
     if (diffMins < 60) return `in ${diffMins}m`;
--- a/src/client/pages/History.tsx
+++ b/src/client/pages/History.tsx
@@ -22,6 +22,11 @@ function stripKindPrefix(summary: string): string {
   return summary.replace(/^(RSS sync|Full sync|Manual sync|Collection publish|Collection sync)[:\s]*/i, "").trim();
 }
 
+function capitalizeSentence(text: string): string {
+  if (!text) return text;
+  return text.charAt(0).toUpperCase() + text.slice(1);
+}
+
 const ACTION_LABELS: Record<string, string> = {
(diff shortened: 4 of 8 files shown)


Migz93:
## Code Review Findings

A few issues to address before merge:

### 1. Duplicate `capitalizeSentence` / `titleCaseStatus` functions

`History.tsx` defines both `capitalizeSentence` and `titleCaseStatus` — they are identical implementations. `titleCaseStatus` should call `capitalizeSentence` instead of duplicating the logic.

### 2. `extraGenericSuccesses` is a dead path for `full`/`user` runs

In `groupSuccessfulSteps`, the `else` block already iterates all success items except `watchlist.fetch/rss/rss.self`. The `extraGenericSuccesses` filter at the bottom then applies an allow-list of `collection.publish.followup | isolation.filters` for the same kinds — both of which were already added by the `else` block. The dedup check silently prevents double-adding so the behaviour is correct, but the extra block is a no-op for `full`/`user` runs and makes the function harder to reason about. This should be simplified/removed for those kinds.

### 3. `collectErrorSteps` dedup by string equality is fragile

```ts
if (item.action === "collection.publish" && run.error && meta && run.error === meta) {
  continue;
}
```

A trailing period, whitespace difference, or any minor formatting divergence between how the error is stored at the run level vs. the item level would break this and cause duplicate error entries. This dedup logic needs to be made more robust.

### 4. `collectionName` is recorded server-side but not surfaced in the UI

`services.ts` now records `collectionName` on `collection.publish` success items, but `groupSuccessfulSteps` drops it. The label should be updated to include the collection name, e.g.:

> Published Movies collection "My Movies" for Miguel

So the format becomes: `Published {mediaType} collection "{collectionName}" for {displayName}`.

Migz93:
Reviewed the follow-up commit (a5e1ed1). All five findings are addressed — agree with the changes, looks good to merge.

PR #93, opened 2026-04-16
fix: reconcile stale running sync_runs on startup

### LINKED ISSUE #89: bug: jobs can remain stuck in Running state indefinitely
Some collection sync jobs were started while onboarding was still in progress, which meant they could not actually complete successfully. Those jobs are now stuck showing as `Running` indefinitely instead of failing, timing out, or being marked as aborted.

This suggests Hubarr's job lifecycle handling is too optimistic when prerequisites are missing or a job becomes unable to complete.

Observed behaviour:
- Collection syncs were kicked off before onboarding was fully complete
- The jobs could not actually run correctly
- They remained in the `Running` state instead of transitioning to a failed/aborted state

Expected behaviour:
- If a job cannot proceed because a dependency or prerequisite is missing, it should fail cleanly
- If a job loses access to something important such as Plex, it should not remain `Running` forever
- Long-running jobs should eventually time out or error rather than staying active indefinitely
- After an application restart, any previously `Running` jobs should be reconciled and marked as errored, aborted, or otherwise no longer active

Things to investigate:
- Whether onboarding can trigger jobs before the system is actually ready for them
- Whether collection sync and other jobs correctly handle missing Plex connectivity or other unmet prerequisites
- Whether jobs have any timeout, heartbeat, or stale-run detection
- Whether job state should be reconciled on startup so stale `Running` entries are not preserved forever
- Whether this issue affects other job types beyond collection sync

The goal is to make job execution more robust so Hubarr does not leave impossible or abandoned tasks permanently stuck in a `Running` state.

Migz93:
## Summary

On startup, any `sync_runs` rows still in `running` status are artefacts of a previous process that crashed or was killed mid-job. Without a reconciliation step these rows remain `running` indefinitely — they can never transition because the process that would have called `completeSyncRun` is gone.

This adds a startup reconciliation pass that immediately marks all orphaned `running` rows as `error` with a clear message (`Aborted: process restarted while job was running`), so the history UI never shows permanently stuck jobs.

Closes #89

## Changes

- `src/server/db/sync.ts` — adds `reconcileStaleRuns(db)` which issues a single `UPDATE` to flip every `running` row to `error` with a `completed_at` timestamp and a descriptive error message.
- `src/server/db/index.ts` — calls `syncRepo.reconcileStaleRuns(this.db)` in the `HubarrDatabase` constructor, immediately after migrations run, so reconciliation happens on every process start before any jobs are registered or scheduled.

## Test plan

- [ ] Start the app normally, trigger a full sync, confirm it completes and shows `success` in the history.
- [ ] Start a sync, kill the process mid-run, restart — confirm the previously `running` row now shows `error` with the message `Aborted: process restarted while job was running`.
- [ ] Confirm no `running` rows remain in the history after startup under any normal condition.
- [ ] Confirm newly started jobs still transition through `running → success/error` correctly after the fix.

🤖 Generated with Claude Code

## Summary by CodeRabbit

* **Bug Fixes**
  * Automatically mark interrupted sync jobs as failed during startup to prevent them from remaining stuck in a running state.
  * Emit warnings when interrupted jobs are found and reconciled, and debug messages when none are found.
  * Improved startup logging around the database recovery process.

### DIFF
--- a/src/server/app.ts
+++ b/src/server/app.ts
@@ -48,7 +48,7 @@ function signedValue(secret: string, value: string) {
 
 export function createApp(config: RuntimeConfig, scheduler?: JobScheduler) {
   const logger = new Logger(config.dataDir);
-  const db = new HubarrDatabase(config);
+  const db = new HubarrDatabase(config, logger);
   const sessionSecret = db.getSessionSecret();
   const imageCache = new ImageCacheService(config.dataDir, db, logger);
   const services = new HubarrServices(db, logger, imageCache);
--- a/src/server/db/index.ts
+++ b/src/server/db/index.ts
@@ -27,17 +27,20 @@ import * as settingsRepo from "./settings.js";
 import * as syncRepo from "./sync.js";
 import * as usersRepo from "./users.js";
 import * as watchlistRepo from "./watchlist.js";
+import type { Logger } from "../logger.js";
 
 export class HubarrDatabase {
   private readonly db: Database.Database;
   private readonly sessionSecret: string;
 
-  constructor(config: RuntimeConfig) {
+  constructor(config: RuntimeConfig, logger?: Logger) {
     this.db = new Database(path.join(config.dataDir, "hubarr.db"));
     this.db.pragma("journal_mode = WAL");
     this.db.pragma("foreign_keys = ON");
     runMigrations(this.db);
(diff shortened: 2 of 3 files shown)


Migz93:
@coderabbit review

PR #106, opened 2026-04-19
chore: improve logging coverage across server workflows

### LINKED ISSUE #45: enhancement: review logging coverage across the existing codebase
We should review the current codebase to make sure logging is present everywhere it is relevant and useful.

At the moment, some areas may already be well covered while others may have little or no logging, which can make debugging, support, and general visibility harder than it needs to be.

This should be treated as a codebase-wide review rather than a change limited to one feature.

Goals:
- Identify parts of the application where logging is missing or too minimal
- Review whether important workflows, background processes, integrations, and error paths are adequately logged
- Improve consistency in how logging is used across the project
- Make sure logs are useful for understanding application behaviour without becoming noisy or excessive

The outcome should be a clearer picture of current logging coverage, along with improvements in the areas where additional logging would provide the most value.

Migz93:
Closes #45

## Summary

This PR expands logging coverage across the existing server codebase so the main startup, scheduling, Plex integration, and admin-driven workflows are easier to follow in production. The goal was not just to add more logs, but to add the missing state-transition and failure-context logs that make debugging background jobs and setup/runtime behaviour much easier without adding duplicate noise.

## Changes

- Updated the scheduler lifecycle so job registration, persisted state hydration, manual triggers, config changes, scheduling decisions, and execution outcomes are all logged in one place.
- Extended service-layer logging for database migration flow, Plex user discovery, Plex token refresh, Plex availability scans, RSS initialization, collection reset, and Plex collection operations so important background work is visible end to end.
- Added request-level logging where it provides useful operational context, including Plex configuration saves, auth failures, onboarding completion, user updates, and application settings changes, while trimming duplicate app-layer logs where the scheduler or service layer already records the same action.
- Wired the scheduler up to the application logger at startup so the new lifecycle logs are emitted consistently during normal runtime.

## Test plan

- [x] Run `npm run check`
- [x] Run `npm run build`
- [x] Review the final logging pass to keep high-signal `info`/`warn` logs and move duplicate/too-noisy additions out of the app layer

🤖 Generated with Codex

## Summary by CodeRabbit

* **Chores**
  * Expanded structured logging and diagnostics across Plex integration, job scheduling, database migrations, onboarding, and sync workflows for better observability.

* **Improvements**
  * User profile updates now enforce allowed fields and validate values before applying changes; updates are logged with field summaries.
  * Settings updates produce deterministic summaries when changed; manual job runs (activity cache) and onboarding steps emit info logs.
  * Availability scans skip when no libraries are configured.

* **Stability**
  * Plex setup/auth and library operations surface warnings/info on failures and owner mismatches.

### DIFF
--- a/src/server/app.ts
+++ b/src/server/app.ts
@@ -11,7 +11,8 @@ import type {
   PlexConnectionOption,
   SettingsResponse,
   SessionUser,
-  SetupStatusResponse
+  SetupStatusResponse,
+  VisibilityConfig
 } from "../shared/types.js";
 import { createSessionId } from "./auth.js";
 import type { RuntimeConfig } from "./config.js";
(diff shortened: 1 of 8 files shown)


Migz93:
I reviewed the latest CodeRabbit note about adding `logger.info("Manual job run requested", { jobId })` before every `scheduler.runNow(...)` call, and I don't think an extra change is needed here.

For scheduler-backed jobs, manual trigger visibility already exists centrally in `JobScheduler.runNow()` via the `"Job triggered manually"` log, so adding the same signal again in the API handler would just duplicate the event at a second layer. The `activity-cache-fetch` branch is the one exception because it bypasses the scheduler and calls `services.syncActivityCache()` directly, which is why the route-level manual-trigger log is still kept there.

I'm leaving this as-is so we keep one clear log entry per scheduler-backed manual trigger rather than two near-identical ones.

🤖 Generated with Codex

PR #119, opened 2026-04-22
fix: use stable username-based labels and label-first collection lookup

### LINKED ISSUE #114: bug: fresh setup can leave duplicate Plex collections after display name override changes
A fresh Hubarr setup can leave behind duplicate Plex collections when users previously had display name overrides.

Scenario observed:
- Hubarr was set up with two users
- Both users had display name overrides
- Their Plex collections were created using the overridden display names
- Hubarr config was then deleted and onboarding was run again from scratch
- The same two users were enabled during onboarding
- Hubarr created new collections using the users' original Plex names
- The old collections with the overridden display names were left behind in Plex

Result:
There are now duplicate/stale collections in Plex for the same users: the old collections using the display name overrides, and the new collections using the original user names.

This needs investigation because Hubarr should be able to identify and reconcile existing user collections even if the display name or collection name has changed.

Things to investigate:
- Whether changing a user's display name currently checks for an existing collection with the old name and renames it
- Whether Hubarr collection labels are currently based on the display name override instead of a stable Plex username or user identifier
- Whether collection labels should always use a stable user identity, such as the full Plex username, rather than the display name override
- Whether Hubarr can reliably find existing collections for a user based on its Hubarr label, even if the collection name is different from the expected current name
- Whether stale/different-named collections should be renamed or deleted when Hubarr creates or reconciles collections

Preferred behaviour:
- Hubarr should not leave stale duplicate collections behind when display names or collection names change
- Hubarr should use a stable way to identify which collections belong to which user
- If Hubarr finds an existing collection for the same user under an old/different name, it should prefer renaming it to the current expected name rather than creating a duplicate
- If multiple stale collections exist for the same user, Hubarr should be able to cleanly reconcile them, either by renaming the correct one or removing extras where appropriate

The goal is for Hubarr to always be able to find, rename, or clean up a user's existing collections based on a stable Hubarr identity/label, rather than relying only on the current display name or collection name.

Migz93:
## Summary

Fixes a bug where Hubarr would create duplicate Plex collections when display names changed or a fresh setup was run without the same display name overrides. The root cause was two compounding issues: collection labels used the mutable `displayName` (which reflects any override) instead of the stable Plex `username`, and collection lookup only searched by title rather than by label.

With this fix, each user's Hubarr label is permanently tied to their Plex username. When `ensureCollection` runs, it first checks by title (fast path for the normal case), then scans all collections in the library for the user's stable label. If it finds a match under an old or different name it renames the collection rather than creating a duplicate. Isolation filters follow the same change and are now keyed to `username` too.

Closes #114

## Changes

- **`src/server/integrations/plex.ts`**
  - `ensureCollection` now accepts `username` and performs label-first lookup after the title check fails. Found collections with a mismatched title are renamed to the current expected name before being returned.
  - New `updateCollectionTitle` helper to rename a Plex collection via the metadata API.
  - `syncIsolationFilters` now generates labels from `u.username` instead of `u.displayName` at both the allLabels build and the per-user exclusion filter.
  - `createCollectionLabel` is unchanged — it still normalises any string passed to it; callers now pass `username`.

- **`src/server/services.ts`**
  - `publishUserCollections`: passes `friend.username` to both `ensureCollection` and `createCollectionLabel`.

## Test plan

- [x] Open Settings → Users, set a display name override for a user, trigger a sync. Confirm the existing collection is renamed in Plex to the new name rather than a second collection being created.
- [x] Delete the Hubarr config, run onboarding fresh. Confirm Hubarr finds and reconciles the existing (now username-labelled) collections rather than creating duplicates.
- [x] Confirm isolation filters still work: each shared Plex user should only see their own watchlist collection, not others'.
- [x] Confirm the Plex collection label is now `hubarr:{plex-username}-watchlist` rather than `hubarr:{display-name-override}-watchlist`.

🤖 Generated with Claude Code

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate Plex collections by reconciling and consolidating per-user collections; extras are removed and the canonical collection is renamed and locked to the expected title.
  * Made label reads more resilient by skipping failed label fetches instead of aborting reconciliation.
  * Fixed isolation label logic to use each user's account identifier (username) and updated exclusion handling accordingly.
* **New Features**
  * Creates a collection when no suitable canonical collection exists.

### DIFF
--- a/src/server/integrations/plex.ts
+++ b/src/server/integrations/plex.ts
@@ -1296,13 +1296,32 @@ export class PlexIntegration {
     const collections = await this.getCollections(libraryId);
     const label = this.createCollectionLabel(username);
 
-    // Fetch all collection labels in parallel so we can check for both title
-    // matches and label matches in one pass without serialising N round-trips.
-    const withLabels = await Promise.all(
+    // Fetch all collection labels in parallel. Use allSettled so a single
+    // failing label fetch (network blip, permission issue) is logged and skipped
+    // rather than aborting the entire reconciliation.
+    const settled = await Promise.allSettled(
       collections.map(async (c) => ({ ...c, labels: await this.getCollectionLabels(c.ratingKey) }))
     );
+    const withLabels = settled
+      .map((result, i) => {
+        if (result.status === "fulfilled") return result.value;
+        this.logger.warn("Failed to fetch labels for collection, skipping in reconciliation", {
+          ratingKey: collections[i].ratingKey,
+          title: collections[i].title,
+          error: result.reason instanceof Error ? result.reason.message : String(result.reason)
+        });
+        return null;
+      })
+      .filter((c): c is { ratingKey: string; title: string; labels: string[] } => c !== null);
 
-    const byTitle = withLabels.find((c) => c.title === title);
+    // Only accept a title match that already carries this user's label or has no
+    // hubarr:* label at all — this avoids accidentally claiming another user's
+    // same-named collection.
+    const byTitle = withLabels.find(
+      (c) => c.title === title &&
+             (c.labels.some((l) => l.toLowerCase() === label.toLowerCase()) ||
+              !c.labels.some((l) => l.toLowerCase().startsWith("hubarr:")))
+    );
     // Label matches on collections other than the title match are duplicates to reconcile.
     const labelMatches = withLabels.filter(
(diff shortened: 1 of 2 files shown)


Migz93:
@coderabbitai review

Migz93:
@coderabbitai review

PR #122, opened 2026-04-23
fix: flatten legacy migrations and remove pre-V2 compatibility code

### LINKED ISSUE #72: enhancement: flatten legacy migrations before next main release
## Summary

Before the next release to `main`, we should flatten the current migration history so the schema and configuration shape we have already migrated to becomes the default baseline for fresh installs.

Right now the codebase still carries transitional migration logic intended to preserve older states during development. That is useful while changes are in flight, but it should not remain in place once we cut the next breaking release. At that point, new installs should start directly from the final shape, and upgrade-only compatibility code should be removed.

## Why this matters

- It keeps startup and persistence logic simpler and easier to reason about.
- It avoids shipping one-off migration branches that only exist to bridge pre-release states.
- It makes the next `main` release a cleaner reset point for future development.

## Code to revisit

The session secret migration added in [src/server/db/settings.ts](src/server/db/settings.ts) is a good example of the kind of transitional code we should remove before release. In particular, `resolveSessionSecret()` currently contains one-time migration behavior for the legacy `.session_secret` file:

- `src/server/db/settings.ts:75`
- `src/server/db/settings.ts:86`
- `src/server/db/settings.ts:90`
- `src/server/db/settings.ts:97`

Once we flatten the baseline, this path should become the default behavior instead of carrying legacy file-to-database migration logic.

## Expected outcome

Before the next `main` release:

- remove transitional migration code that only supports pre-release states
- make the migrated-to schema/config layout the default baseline
- update any docs or PR descriptions that currently imply upgrade compatibility we no longer intend to preserve

🤖 Generated with Codex

Migz93:
## Summary

Implements the V2 migration flatten described in issue #72. V2 is a clean-slate release — no prior installs need to be supported — so all transitional compatibility code that existed solely to bridge pre-release states to the new schema/config layout has been removed. Fresh installs now start directly from the final shape.

## Changes

**`src/server/db/migrations.ts`**
- Removed `inferSchemaVersion()`, `v1Tables`, and `tableExists()` — these handled DBs that predated the version-tracking system; irrelevant for V2 fresh installs
- Removed the schema version inference block from `runMigrations()` — `user_version = 0` now simply means a new empty DB
- Stripped the RSS stableKey-to-plex:// normalisation JS loop from migration v7 (the `ALTER TABLE` and `UPDATE` SQL that defines the column are kept)
- Stripped the watchlist/user backfill JS loop from migration v8 (the four `CREATE TABLE`/`CREATE INDEX` statements that define the schema are kept)
- Removed the three private identifier helper functions (`normalizeIdentifierValue`, `inferUserIdentifierType`, `inferMediaIdentifierType`) that were only used by the now-deleted v8 backfill loop
- Deleted migration v11 entirely; updated migration v10 to define `seerr_user_links.auto_request_enabled` as nullable (`INTEGER`) directly, removing the need for the rename/recreate dance v11 performed

**`src/server/db/settings.ts`**
- Removed `normalizeSortOrder()` and its call in `getAppSettings()` — transparently rewrote stored `year-desc`/`year-asc` values on every read; V2 default is already `date-desc`
- Simplified `resolveSessionSecret()`: removed the legacy `.session_secret` file-read path, the `dataDir` parameter, and the now-unused `fs`/`path` imports
- Removed the `onboardingComplete`-from-`configurationValid` backwards-compat derivation in `getBootstrapStatus()` — V2 installs always have the flag

**`src/server/db/index.ts`**
- Updated `resolveSessionSecret` call to remove the dropped `dataDir` argument

**`src/server/app.ts`**
- Removed `year-desc` and `year-asc` from the watchlist `sortBy` query param allowlist
- Replaced the `legacyMap` translation block in the settings PATCH handler with a direct validity check against the canonical sort order values

**`src/server/db/watchlist.ts`**
- Removed the `raw_payload` fallback for `discoverKey` on rows written before migration v7 — all V2 rows will have the dedicated column populated

**`src/server/db/identifiers.ts` / `src/server/db/seerr.ts`**
- Exported `normalizeIdentifierValue` from `identifiers.ts` as the single canonical implementation
- Removed the private duplicate from `seerr.ts` and replaced its usages with the import

## Test plan

- [ ] Build the image locally (`docker build -t hubarr .`) and confirm it compiles cleanly
- [ ] Start a container against a fresh empty config dir (no existing DB) and complete onboarding from scratch — confirm all steps work and the app reaches a running state normally
- [ ] Confirm the existing live container (with carry-over DB) still starts and runs correctly — migrations v1–v10 should apply as normal on any DB at version < 10

Closes #72

🤖 Generated with Claude Code

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed "year-asc" and "year-desc" sorting on watchlists; unsupported sort values now fall back to "added-desc".

* **Refactor**
  * Centralized identifier normalization for more consistent item matching.
  * Watchlist ordering simplified to rely on the primary added-date and shared secondary comparators.

* **Documentation**
  * Migration docs updated to reflect a flattened baseline schema and guidance for future migrations.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -45,10 +45,12 @@ Hubarr uses SQLite `PRAGMA user_version` for schema migrations.
 - `runMigrations(db)` runs on startup, applies any migration whose version is higher than the current `user_version`, and advances `user_version` after each successful migration
 - Each migration runs inside a transaction so a failure should leave the database unchanged
 
+**V2 baseline:** the schema was flattened at the V2 release. Migration version 1 is a single `CREATE TABLE` block that defines every table and index in its final form. A fresh install runs it once and is immediately at the correct state — there are no intermediate ALTER TABLE steps or sequential migrations to follow.
+
 When changing the schema in the future:
 
 1. Add a new migration entry with the next integer version
-2. Write the schema change in that migration's `up(db)` function
+2. Write the schema change in that migration's `up(db)` function (ALTER TABLE, CREATE TABLE, etc.)
 3. Do not edit older migrations that may already have shipped
 4. Keep default-setting seeding separate from schema migrations
 
--- a/src/server/app.ts
+++ b/src/server/app.ts
@@ -696,8 +696,7 @@ export function createApp(config: RuntimeConfig, scheduler?: JobScheduler) {
     const sortByParam = req.query["sortBy"];
     const sortBy =
       sortByParam === "added-desc" || sortByParam === "added-asc" ||
-      sortByParam === "title-asc" || sortByParam === "title-desc" ||
-      sortByParam === "year-desc" || sortByParam === "year-asc"
+      sortByParam === "title-asc" || sortByParam === "title-desc"
         ? sortByParam
         : "added-desc";
     const page = req.query["page"] ? Math.max(1, Number(req.query["page"])) : 1;
--- a/src/server/db/identifiers.ts
+++ b/src/server/db/identifiers.ts
@@ -4,7 +4,7 @@ import type { WatchlistItem } from "../../shared/types.js";
 type MediaIdentifierType = "plex_guid" | "discover_key" | "imdb" | "tmdb" | "tvdb" | "guid";
 type UserIdentifierType = "plex_user" | "plex_numeric" | "plex_uuid";
 
-function normalizeIdentifierValue(value: string): string {
+export function normalizeIdentifierValue(value: string): string {
   return value.trim().toLowerCase();
 }
 
--- a/src/server/db/migrations.ts
+++ b/src/server/db/migrations.ts
@@ -6,26 +6,6 @@ interface Migration {
   up(db: Database.Database): void;
 }
 
-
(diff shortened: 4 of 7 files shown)


Migz93:
**Re: CodeRabbit finding — migration flattening removes upgrade path**

Intentional — won't fix.

This is a deliberate V2 design decision. Hubarr currently has a single active install (the maintainer's own instance) and V2 is an explicit clean-break release where backward compatibility with pre-V2 databases is not a goal. The database is wiped as part of the V2 upgrade, so there are no carry-over installs to protect.

The flattened single migration is the correct approach for this project at this point in time:

- Fresh V2 installs get the complete final schema in one step
- `user_version` is still tracked correctly (0 → 1)
- There is no historical upgrade path to preserve because none is needed

If Hubarr ever reaches a point where multiple users are running different versions simultaneously, incremental migrations from a stable baseline would be the right approach. That point hasn't been reached yet.

🤖 Generated with Claude Code

PR #123, opened 2026-04-23
fix: harden watchlist query parsing

Migz93:
## Summary

This PR hardens the `/api/watchlists` query parsing so malformed numeric parameters no longer propagate `NaN` into pagination and filtering logic. Valid requests keep the existing behavior, while invalid `userId`, `page`, and `pageSize` values now fall back to safe defaults.

## Changes

- Added a small `parsePositiveIntegerQuery()` helper in `src/server/app.ts` to centralize positive-integer parsing for query string values.
- Updated the `/api/watchlists` route to use the helper for `userId`, `page`, and `pageSize` instead of calling `Number(...)` inline.
- Preserved the existing defaults and `pageSize` cap while ensuring malformed values are ignored cleanly.

## Test plan

- [x] Run `npm run build`
- [ ] Call `GET /api/watchlists` with valid `userId`, `page`, and `pageSize` values and confirm the response matches existing behavior.
- [ ] Call `GET /api/watchlists` with malformed values such as `page=abc`, `pageSize=-1`, or `userId=` and confirm the endpoint falls back to safe defaults instead of producing invalid pagination state.

🤖 Generated with Codex

## Summary by CodeRabbit

* **Bug Fixes**
  * Strict validation for watchlists API query parameters: positive integers parsed correctly; invalid or empty values now fall back to sensible defaults (page → 1, pageSize → 50; userId → undefined).
  * pageSize is capped at a maximum of 200 to prevent oversized requests.

* **New Features**
  * Conditional debug logging when parameters are invalid or are clamped, including raw and effective values.

### DIFF
--- a/src/server/app.ts
+++ b/src/server/app.ts
@@ -46,6 +46,19 @@ function parseCookies(rawCookie = "") {
     }, new Map());
 }
 
+function parsePositiveIntegerQuery(value: unknown): number | undefined {
+  if (typeof value !== "string" || value.trim() === "") {
+    return undefined;
+  }
+
+  const parsed = Number(value);
+  if (!Number.isInteger(parsed) || parsed < 1) {
+    return undefined;
+  }
+
+  return parsed;
+}
+
 function signedValue(secret: string, value: string) {
   return crypto.createHmac("sha256", secret).update(value).digest("hex");
 }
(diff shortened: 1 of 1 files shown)


Migz93:
@coderabbit review

PR #244, opened 2026-07-31
fix: run container safely with writable bind mounts

### LINKED ISSUE #241: Run the production container as non-root safely against fresh bind mounts
## Summary

Hubarr's production image runs as `node` (non-root) via a build-time `chown`, but this doesn't hold up against the exact volume setup the README tells users to use — a host bind mount. A genuinely fresh install can fail to start.

## Where this comes from

Current `Dockerfile` (runtime stage):

```dockerfile
COPY --chown=node:node --from=build /app/package.json ./package.json
COPY --chown=node:node --from=production-deps /app/node_modules ./node_modules
COPY --chown=node:node --from=build /app/dist ./dist
RUN mkdir -p /config && chown node:node /config
USER node
```

The README's Docker instructions use a host bind mount:

```bash
docker run -d --name hubarr -p 9301:9301 -v /opt/hubarr:/config ghcr.io/migz93/hubarr:latest
```

`chown node:node /config` at build time only affects the directory *baked into the image*. A host bind mount replaces that directory entirely at container start — it does not inherit anything from the image. If `/opt/hubarr` doesn't already exist on the host, Docker creates it there as `root:root` before the container's `USER node` ever takes effect, and the `node` user then can't write to it.

Reproduced this exact class of failure while fixing the same issue in ShelfBridge (`SqliteError: unable to open database file` / `SQLITE_CANTOPEN`) against a fresh, empty bind-mounted volume with the equivalent Dockerfile shape.

Note: this is specifically about a *host bind mount that doesn't pre-exist*. If Hubarr is ever run against a named Docker volume instead, the image's baked-in `/config` ownership would actually apply (named volumes copy the image directory's content/ownership on first creation) — but that's not what the documented Docker Compose / `docker run` examples use.

## How I fixed it in ShelfBridge

Reference implementation: [Migz93/shelfbridge#45](https://github.com/Migz93/shelfbridge/pull/45)

1. Added a small entrypoint script (`docker-entrypoint.sh`) instead of relying on a build-time `chown`:

   ```sh
   #!/bin/sh
   set -e

   if [ "$(id -u)" = "0" ]; then
     mkdir -p "$DATA_DIR"
     # Only pay for a recursive chown when the top-level directory isn't
     # already node-owned — DATA_DIR can grow large (image cache, artwork,
     # etc.), so re-walking it on every restart once ownership is already
     # correct would slow startup for no benefit.
     current_owner="$(stat -c '%u' "$DATA_DIR")"
     node_uid="$(id -u node)"
     if [ "$current_owner" != "$node_uid" ]; then
       chown -R node:node "$DATA_DIR"
     fi
     exec gosu node "$@"
   fi

   exec "$@"
   ```

   (ShelfBridge is Alpine-based and uses `su-exec`; Hubarr is `node:22-trixie-slim`/Debian-based, so use `gosu` instead — same idea, `apt-get install -y gosu`.)

2. Dockerfile changes: container now starts as root only so the entrypoint can fix `DATA_DIR` ownership, then drops to `node` via `gosu`/`su-exec` before `exec`'ing the real command:

   ```dockerfile
   RUN apt-get update && apt-get install -y --no-install-recommends gosu && rm -rf /var/lib/apt/lists/*
   COPY docker-entrypoint.sh /entrypoint.sh
   RUN chmod +x /entrypoint.sh
   ENTRYPOINT ["/entrypoint.sh"]
   CMD ["node", "dist/server/server/index.js"]
   ```

3. **Also removed `--chown=node:node` from the app code `COPY` layers.** A code-review pass (Codex) on the ShelfBridge PR flagged that the app process only ever needs to *read* its own code/dependencies — giving the runtime user ownership of `/app` means a compromised app could rewrite its own code/dependencies from inside the container, which weakens the point of running non-root in the first place. `/app` should stay root-owned and merely world-readable; only `/config` needs to be writable by `node`. Hubarr's current Dockerfile has this same pattern (`COPY --chown=node:node ... /app/...`) and should drop the `--chown` there too.

## What to change here

- [ ] Add `gosu` and an entrypoint script that starts as root, conditionally `chown -R`s `DATA_DIR` (only when the top-level owner doesn't already match `node`), then drops to `node` via `gosu` before exec'ing `CMD`.
- [ ] Remove `--chown=node:node` from the `COPY` lines for `/app/package.json`, `/app/node_modules`, `/app/dist` — leave them root-owned.
- [ ] Keep (or drop, now redundant) `RUN mkdir -p /config && chown node:node /config` — harmless as a fallback default, but no longer load-bearing once the entrypoint handles it.
- [ ] Test against: a genuinely fresh empty host bind mount (this is the case that currently breaks), an existing data directory from an older root-run container, and confirm via `docker exec <container> ps aux` that the actual `node` process (PID 1) runs as `node`, not `root`.

## Acceptance criteria (mirroring the intent of the ShelfBridge fix)

- `docker exec <container> ps aux` shows the app process running as a non-root UID.
- A brand-new, empty `/opt/hubarr`-style host bind mount works on first `docker run` — no manual `chown` required on the host.
- An existing data directory (however it was previously owned) still works after upgrading to this image.
- `/app` stays root-owned; only `/config` (and its contents) end up owned by the runtime user.
- No privileged mode or added Linux capabilities required.

🤖 Generated with Claude Code

Migz93:
## Summary

Fix production container startup on fresh host bind mounts and when upgrading existing root-owned data directories. The image now starts with the privileges needed to repair `/config`, then runs the Hubarr process as the non-root `node` user. This removes the need for manual host-side `chown` while preserving the container’s least-privilege boundary.

## Changes

- `Dockerfile`: install Debian `gosu`, keep `/app` files root-owned, add the entrypoint, and start the container through it instead of setting `USER node` at build time.
- `docker-entrypoint.sh`: create `/config` when needed, repair mismatched ownership throughout the data tree, use `chown -h` to prevent symlink targets outside `/config` from being modified, and `exec` the application as `node`.
- The entrypoint preserves explicit `--user 1000:1000` operation by directly executing the command when it is not running as root.

## Test plan

- [x] Run `npm run check`, `npm run lint`, and `npm run build`.
- [x] Build the Docker image and verify PID 1 runs as UID/GID 1000 while `/app` remains `root:root` and `/config` is writable by `node`.
- [x] Test a fresh empty bind mount and confirm Hubarr starts without manual host-side ownership changes.
- [x] Test an existing root-owned data directory and confirm ownership is repaired on startup.
- [x] Test nested ownership mismatches, an explicit `--user 1000:1000` override, and a symlink escaping `/config`; confirm the external target remains unchanged.
- [x] Recreate the local `hubarr` container from this image using `/opt/hubarr:/config` and confirm normal operation.

Closes #241

🤖 Generated with Codex

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container startup by automatically preparing data directory permissions.
  * Ensured applications run with appropriate non-root privileges when started as root.
  * Added safer handling for ownership changes, including symbolic links and inaccessible files.
  * Preserved direct command execution when the container already runs as a non-root user.
  * Added validation for data directory paths to prevent unsafe or invalid locations.
  * Improved reliability when launching containers with existing or partially inaccessible data.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,12 +31,13 @@ ARG COMMIT_SHA=local
 ENV BUILD_CHANNEL=$BUILD_CHANNEL
 ENV COMMIT_SHA=$COMMIT_SHA
 RUN apt-get update \
-  && apt-get install -y --no-install-recommends fontconfig fonts-dejavu-core gosu \
+  && apt-get install -y --no-install-recommends fontconfig fonts-dejavu-core gosu python3 \
   && rm -rf /var/lib/apt/lists/*
 COPY --from=build /app/package.json ./package.json
 COPY --from=production-deps /app/node_modules ./node_modules
 COPY --from=build /app/dist ./dist
 COPY docker-entrypoint.sh /entrypoint.sh
+COPY docker-ownership-repair.py /ownership-repair.py
 RUN chmod 755 /entrypoint.sh
 RUN mkdir -p /config && chown node:node /config
 ENTRYPOINT ["/entrypoint.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,12 +31,14 @@ ARG COMMIT_SHA=local
 ENV BUILD_CHANNEL=$BUILD_CHANNEL
 ENV COMMIT_SHA=$COMMIT_SHA
 RUN apt-get update \
-  && apt-get install -y --no-install-recommends fontconfig fonts-dejavu-core \
+  && apt-get install -y --no-install-recommends fontconfig fonts-dejavu-core gosu \
   && rm -rf /var/lib/apt/lists/*
-COPY --chown=node:node --from=build /app/package.json ./package.json
-COPY --chown=node:node --from=production-deps /app/node_modules ./node_modules
-COPY --chown=node:node --from=build /app/dist ./dist
+COPY --from=build /app/package.json ./package.json
+COPY --from=production-deps /app/node_modules ./node_modules
+COPY --from=build /app/dist ./dist
+COPY docker-entrypoint.sh /entrypoint.sh
+RUN chmod 755 /entrypoint.sh
 RUN mkdir -p /config && chown node:node /config
-USER node
+ENTRYPOINT ["/entrypoint.sh"]
 EXPOSE 9301
 CMD ["node", "dist/server/server/index.js"]
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+set -e
+
+if [ "$(id -u)" = "0" ]; then
+  mkdir -p "$DATA_DIR"
+
+  # DATA_DIR may be a fresh host bind mount, or contain files restored or
+  # copied in with different ownership after the directory was initialized.
+  # Walk the tree so nested mismatches are repaired, but only chown entries
+  # that need it and never follow symlinks into files outside the data
+  # directory.
+  find "$DATA_DIR" \( ! -user node -o ! -group node \) -exec chown -h node:node {} +
+
+  exec gosu node "$@"
+fi
+
+exec "$@"
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -2,14 +2,30 @@
 set -e
 
 if [ "$(id -u)" = "0" ]; then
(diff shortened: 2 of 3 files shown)


Migz93:
@coderabbit review

Migz93:
@coderabbit review

